Issue #23, opened 2020-06-25
Angular_component proto dependency not compatible with protoc version


hanj7221:
The most recent version for angular_component on pub is 13.6. It's proto dependency is old and it breaking compatibility with protoc 3.12. The solution (temporary) is to replace all material buttons with just button.

PR #28, opened 2020-06-26
Creating service classes to handle query submission and parsing

### LINKED ISSUE #5: Create service class that handles query submission and parsing [1 day]
A service class will be needed to store the query information - the predetermined advertiser id and insertion order id for this milestone, but later this will be expanded to also campaign ids and etc. This class should also handle the query submission action and pass the response over to the parser.

Cost estimate: [1 day]

hanj7221:
Closing issue #5 

New issues that arose in this PR and will be resolved in a separate PR:
- Protobuff refactoring 'InsertionOrderBudget" to just "Budget": #30
- Put all service related classes into their own service folder: #31
- Replace singleton pattern with @Injectable(): #32 
- Query service tests: #29


lib/src/credential_component.dart
```diff
@@ -1,17 +1,15 @@
 import 'package:angular/angular.dart';
-import 'package:angular_components/material_button/material_button.dart';
 
 import 'credential_service.dart';
 
 @Component(
   selector: 'credential',
   template: '''
-    <material-button (click)="onClick()" debugId="sign-on-btn">
```
ruben-aghayan, line 9: button refactor should be its own pr

lib/src/credential_component.dart
```diff
@@ -20,9 +18,7 @@ class CredentialComponent implements OnInit {
   CredentialComponent(this._credential);
 
   @override
-  void ngOnInit() {
-    _credential.handleClientLoad();
-  }
+  void ngOnInit() => _credential.handleClientLoad();
```
ruben-aghayan, line 21: nice

lib/src/parser.dart
```diff
@@ -0,0 +1,222 @@
+import 'dart:convert';
+
+import 'proto/insertion_order.pb.dart';
+
+class Parser {
```
ruben-aghayan: what does this parse? consider making the class name more specific

lib/src/parser.dart
```diff
@@ -0,0 +1,222 @@
...
+    return createInsertionOrder(map);
+  }
+
+  /// Creates an [InsertionOrder] instance from [map].
+  ///
+  /// Throws an [FormatException] if a required field is missing.
+  static InsertionOrder createInsertionOrder(Map<String, dynamic> map) {
+    final insertionOrder = InsertionOrder();
+
+    if (!map.containsKey('advertiserId')) throw FormatException;
```
ruben-aghayan: consider checking for all of these at once
what happens if e.g. advertiser id is missing? can we still show the rest of the information or is it crucial in order to do anything?

lib/src/parser.dart
```diff
@@ -0,0 +1,222 @@
...
+    insertionOrder.advertiserId = map['advertiserId'];
+
+    if (!map.containsKey('campaignId')) throw FormatException;
+    insertionOrder.campaignId = map['campaignId'];
```
ruben-aghayan: are you familiar with cascades? they are an idiomatic way of building a proto
https://dart.dev/guides/language/language-tour#cascade-notation-

lib/src/parser.dart
```diff
@@ -0,0 +1,222 @@
...
+    insertionOrder.insertionOrderId = map['insertionOrderId'];
+
+    if (!map.containsKey('displayName')) throw FormatException;
+    insertionOrder.displayName = map['displayName'];
+
+    if (!map.containsKey('updateTime')) throw FormatException;
+    insertionOrder.updateTime = map['updateTime'];
+
+    if (!map.containsKey('entityStatus')) throw FormatException;
+    insertionOrder.entityStatus = createEntityStatus(map['entityStatus']);
```
ruben-aghayan: really nice breakdown into submethod

lib/src/parser.dart
```diff
@@ -0,0 +1,222 @@
...
+    } else if (map.containsKey('dailyMaxMicros')) {
+      pacing.dailyMaxMicros = map['dailyMaxMicros'];
+    } else if (map.containsKey('dailyMaxImpressions')) {
+      pacing.dailyMaxImpressions = map['dailyMaxImpressions'];
+    }
+
+    return pacing;
+  }
+
+  /// Creates an [InsertionOrder_InsertionOrderBudget] from [map].
```
ruben-aghayan: nice doc comments in this class

lib/src/parser.dart
```diff
@@ -0,0 +1,222 @@
...
+      Map<String, dynamic> map) {
+    final budget = InsertionOrder_InsertionOrderBudget();
+
+    if (!map.containsKey('budgetUnit')) throw FormatException;
+    budget.budgetUnit = createBudgetUnit(map['budgetUnit']);
+
+    budget.automationType = createAutomationType(map['automationType']);
+
+    if (!map.containsKey('budgetSegments')) throw FormatException;
+    for (var map in map['budgetSegments']) {
```
ruben-aghayan: consider another var name since map is already defined

lib/src/parser.dart
```diff
@@ -0,0 +1,222 @@
...
+    }
+    return budget;
+  }
+
+  /// Creates an
+  /// [InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment]
+  /// instance from [map].
+  ///
+  /// Throws an [FormatException] if a required field is missing.
+  static InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment
```
ruben-aghayan: i see what you mean about redundant names. I think we should keep the enums in line with the api but the message here could be named Budget so it would be InsertionOrder_Budget_Segment. wdyt?

lib/src/parser.dart
```diff
@@ -0,0 +1,222 @@
...
+
+  /// Creates an
+  /// [InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment_DateRange]
+  /// instance from [map].
+  static InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment_DateRange
+      createDateRange(Map<String, dynamic> map) {
+    final dateRange =
+        InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment_DateRange();
+
+    if (map.containsKey('startDate')) {
```
ruben-aghayan: did you know you could do if(condition) a.b=c;? if it all fits in one line you can drop the brackets and do it inline.
i find that i mostly use it for escape conditions if(cond) return; so the entire method isnt nested.

lib/src/parser.dart
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  /// Creates an
+  /// [InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment_DateRange_Date]
+  /// from [map].
+  static InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment_DateRange_Date
+      createDate(Map<String, dynamic> map) {
+    final date =
+        InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment_DateRange_Date();
+    if (map.containsKey('year')) date.year = map['year'];
```
ruben-aghayan: oh i see you did it here! nice

lib/src/parser.dart
```diff
@@ -0,0 +1,222 @@
...
+    if (map.containsKey('month')) date.month = map['month'];
+    if (map.containsKey('day')) date.day = map['day'];
+    return date;
+  }
+
+  /// Creates an [InsertionOrder_EntityStatus] enum from [map].
+  ///
+  /// Throws an [FormatException] if [target] doesn't correspond
+  /// to any enum values.
+  static InsertionOrder_EntityStatus createEntityStatus(String target) {
```
ruben-aghayan: can this be private?

lib/src/query_builder.dart
```diff
@@ -0,0 +1,39 @@
+enum QueryType { singleIO }
+
+class QueryBuilder {
```
ruben-aghayan: what does it query? we will have two services so we may benefit if the name is more descriptive.

lib/src/query_service.dart
```diff
@@ -0,0 +1,31 @@
+import 'dart:js';
```
ruben-aghayan, line 1: consider putting all service related classes into their own services folder

test/query_builder_test.dart
```diff
@@ -0,0 +1,38 @@
+import 'dart:math';
+
+import 'package:angular_test/angular_test.dart';
+import 'package:test/test.dart';
+
+import 'package:dv360_excel_plugin/src/query_builder.dart';
+
+void main() {
+  const max = 10000;
```
ruben-aghayan: why is this the max?

test/query_builder_test.dart
```diff
@@ -0,0 +1,38 @@
...
+
+  Random random;
+  QueryBuilder queryBuilder;
+
+  setUp(() {
+    random = Random();
```
ruben-aghayan: i see what you are going for but we typically stay away from randomness in tests because it can cause flakiness issues. Lets think about another way to test the intent here.

test/query_builder_test.dart
```diff
@@ -0,0 +1,38 @@
...
+    queryBuilder = QueryBuilder();
+  });
+
+  tearDown(disposeAnyRunningTest);
+
+  group('generate query for:', () {
```
ruben-aghayan: great use of the test naming system!

test/query_builder_test.dart
```diff
@@ -0,0 +1,38 @@
...
+
+    setUp(() {
+      advertiserId = random.nextInt(max).toString();
+      insertionOrderId = random.nextInt(max).toString();
+      queryBuilder.advertiserId = advertiserId;
+      queryBuilder.insertionOrderId = insertionOrderId;
+    });
+
+    test('a single insertion order', () {
+      final query = 'https://displayvideo.googleapis.com/v1/'
```
ruben-aghayan: we typically structure tests as 

arrange

act

assert

so the setUp would be arrange. expect is assert. the act here is queryBuild.generateQuery()
consider breaking your test up to match this structure. One format could be:

final actual = methodToProduceOutput();

final expected = construction of expected by hand
expect(actual, expected);

test/query_component_test.dart
```diff
@@ -53,13 +58,52 @@ void main() {
...
+    expect(queryBuilder.insertionOrderId, isNull);
+    expect(queryBuilder.advertiserId, isNull);
+  });
+
+  group('update:', () {
+    String id;
+
+    setUp(() async {
+      id = random.nextInt(10000).toString();
+      await queryComponentPO.typeAdvertiserId(id);
```
ruben-aghayan: should be different numbers ideally

ruben-aghayan (changes requested):

lib/src/dv360_query_builder.dart
```diff
@@ -0,0 +1,39 @@
+enum QueryType { singleIO }
```
ruben-aghayan: opt: remove enum. rn it feels like coding ahead

lib/src/dv360_query_builder.dart
```diff
@@ -0,0 +1,39 @@
...
+
+class DV360QueryBuilder {
+  DV360QueryBuilder._private();
+
+  static final DV360QueryBuilder _singleton = DV360QueryBuilder._private();
+
+  factory DV360QueryBuilder() {
```
ruben-aghayan: try removing singleton here, angular di may be singleton by default

lib/src/dv360_query_builder.dart
```diff
@@ -0,0 +1,39 @@
...
+  String _insertionOrderId;
+
+  String get advertiserId => _advertiserId;
+  set advertiserId(String id) => _advertiserId = id;
+
+  String get insertionOrderId => _insertionOrderId;
+  set insertionOrderId(String id) => _insertionOrderId = id;
+
+  /// Generates query based on user selected query type and inputs.
+  String generateQuery() {
```
ruben-aghayan: make advertiser id/ insertion order id args to this method

lib/src/insertion_order_parser.dart
```diff
@@ -0,0 +1,193 @@
+import 'dart:convert';
+
+import 'proto/insertion_order.pb.dart';
+
+class InsertionOrderParser {
+  static const emptyEntry = '';
```
ruben-aghayan: private

lib/src/insertion_order_parser.dart
```diff
@@ -0,0 +1,193 @@
...
+    }
+    return InsertionOrder_Pacing_PacingPeriod.PACING_PERIOD_UNSPECIFIED;
+  }
+
+  /// Creates an [InsertionOrder_Pacing_PacingType] enum from [map].
+  ///
+  /// Returns default value [PACING_TYPE_UNSPECIFIED]
+  /// if [target] doesn't correspond to any enum values.
+  static InsertionOrder_Pacing_PacingType _createPacingType(String target) {
+    for (var type in InsertionOrder_Pacing_PacingType.values) {
```
ruben-aghayan: consider adding types here (and other loops) for readability

lib/src/insertion_order_parser.dart
```diff
@@ -0,0 +1,193 @@
...
+      }
+    }
+    return InsertionOrder_Pacing_PacingType.PACING_TYPE_UNSPECIFIED;
+  }
+
+  /// Creates an [InsertionOrder_InsertionOrderBudget_BudgetUnit] from [map].
+  ///
+  /// Returns default value [BUDGET_UNIT_UNSPECIFIED]
+  /// if [target] doesn't correspond to any enum values.
+  static InsertionOrder_InsertionOrderBudget_BudgetUnit _createBudgetUnit(
```
ruben-aghayan, line 147: is there a better way to parse the string into an enum?
or if not that maybe a static map?

lib/src/query_component.dart
```diff
@@ -1,23 +1,29 @@
...
+  directives: [formDirectives],
 )
 class QueryComponent {
   final buttonName = 'populate';
-  final ExcelDart _excel;
+  final DV360QueryBuilder queryBuilder;
+  final QueryService queryService;
 
-  QueryComponent(this._excel);
+  QueryComponent(this.queryBuilder, this.queryService);
```
ruben-aghayan: where is query builder used? will it ever be used outside query service?

lib/src/query_service.dart
```diff
@@ -0,0 +1,31 @@
...
+  QueryService._private();
+
+  static final QueryService _singleton = QueryService._private();
+
+  factory QueryService() {
+    return _singleton;
+  }
+
+  final queryBuilder = DV360QueryBuilder();
+  final _table = <InsertionOrder>[];
```
ruben-aghayan: nit: why _table? consider another name or adding more descriptors here

test/insertion_order_parser_test.dart
```diff
@@ -0,0 +1,375 @@
+import 'package:angular_test/angular_test.dart';
+import 'package:test/test.dart';
+
+import 'package:dv360_excel_plugin/src/insertion_order_parser.dart';
+import 'package:dv360_excel_plugin/src/proto/insertion_order.pb.dart';
+
+void main() {
+  String input;
```
ruben-aghayan: wrap this all in a group so setup is rerun every time

test/insertion_order_parser_test.dart
```diff
@@ -0,0 +1,375 @@
...
+          ..startDate = oneStartDate
+          ..endDate = oneEndDate;
+  });
+
+  tearDown(disposeAnyRunningTest);
+
+  group('parse insertionOrder from empty json:', () {
+    setUp(() => input = '{}');
+
+    test('function should return normally', () {
```
ruben-aghayan: can delete this test since if it throws "result is an empty insertionOrder instance" will catch it. We by default assume a method should return normally. We do encourage testing when that isnt the case though, e.g. we expect it to throw

test/insertion_order_parser_test.dart
```diff
@@ -0,0 +1,375 @@
...
+      expect(() => InsertionOrderParser.parse(input), returnsNormally);
+    });
+
+    test('result is an insertionOrder with all required fields', () {
+      final actual = InsertionOrderParser.parse(input);
+
+      final pacing = InsertionOrder_Pacing()
+        ..pacingPeriod = InsertionOrder_Pacing_PacingPeriod.PACING_PERIOD_FLIGHT
```
ruben-aghayan, line 115: nice use of cascades

test/insertion_order_parser_test.dart
```diff
@@ -0,0 +1,375 @@
...
+      input = '{$advertiserId, $campaignId, $insertionOrderId, $displayName,'
+          '$entityStatus, $updateTime, $pacing, $budget}';
+    });
+
+    test('function should return normally', () {
+      expect(() => InsertionOrderParser.parse(input), returnsNormally);
+    });
+
+    test('result is an insertionOrder with all required fields', () {
+      final actual = InsertionOrderParser.parse(input);
```
ruben-aghayan, line 112: nice arrange act assert setup

test/insertion_order_parser_test.dart
```diff
@@ -0,0 +1,375 @@
...
+            .BUDGET_UNIT_UNSPECIFIED
+        ..automationType =
+            InsertionOrder_InsertionOrderBudget_InsertionOrderAutomationType
+                .INSERTION_ORDER_AUTOMATION_TYPE_NONE
+        ..budgetSegments.add(segment);
+      final expected = insertionOrderTemplate..budget = budget;
+      expect(actual, expected);
+    });
+
+    test('multiple budgetSegments', () {
```
ruben-aghayan: this is a good test case

ruben-aghayan (changes requested):

lib/src/query_service.dart
hanj7221: Can delete this _table here, not really necessary, and I was coding ahead.

lib/src/insertion_order_parser.dart
hanj7221: aha static map, good point

lib/src/insertion_order_parser.dart
```diff
@@ -0,0 +1,199 @@
...
+import 'proto/insertion_order.pb.dart';
+
+class InsertionOrderParser {
+  static const _emptyEntry = '';
+
+  static final _entityStatusMap = {
+    for (var status in InsertionOrder_EntityStatus.values) status.name: status
+  };
+
+  static final _pacingPeriodMap = {
```
ruben-aghayan: nice use of language features

lib/src/insertion_order_parser.dart
```diff
@@ -0,0 +1,199 @@
...
+  };
+
+  static final _automationTypeMap = {
+    for (var type
+        in InsertionOrder_InsertionOrderBudget_InsertionOrderAutomationType
+            .values)
+      type.name: type
+  };
+
+  /// Parses a json string to [InsertionOrder].
```
ruben-aghayan, line 39: nice formatting on the doc comment to refer to the class!

lib/src/insertion_order_parser.dart
```diff
@@ -0,0 +1,199 @@
...
+    return budget;
+  }
+
+  /// Creates an
+  /// [InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment]
+  /// instance from [map].
+  static InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment
+      _createBudgetSegment(Map<String, dynamic> map) {
+    if (map == null || map.isEmpty) {
+      InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment();
```
ruben-aghayan: did you want to return this?
did a test go through this branch?

lib/src/insertion_order_parser.dart
```diff
@@ -0,0 +1,199 @@
...
+
+    final dateRange =
+        InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment_DateRange()
+          ..startDate = _createDate(map['startDate'])
+          ..endDate = _createDate(map['endDate']);
+
+    return dateRange;
+  }
+
+  /// Creates an
```
ruben-aghayan: opt: documenting private methods is optional so you can remove it if you think the method name is descriptive enough

lib/src/insertion_order_parser.dart
```diff
@@ -0,0 +1,199 @@
...
+    return date;
+  }
+
+  /// Creates an [InsertionOrder_EntityStatus] enum from [map].
+  ///
+  /// Returns default value [ENTITY_STATUS_UNSPECIFIED]
+  /// if [target] doesn't correspond to any enum values.
+  static InsertionOrder_EntityStatus _createEntityStatus(String target) {
+    if (_entityStatusMap.containsKey(target)) return _entityStatusMap[target];
+    return InsertionOrder_EntityStatus.ENTITY_STATUS_UNSPECIFIED;
```
ruben-aghayan: enums have a default value already (whichever has a tag of 0). Can we take advantage of the built-in behavior or do we have to set the default ourselves. The advantage of using the default behavior is .hasX() will return false if returning a value by default.

lib/src/insertion_order_parser.dart
ruben-aghayan: can this method be replaced by something like
_enumMap.get(blah) ?? default val ?
that may be more idiomatic
do you think splitting this out adds to readability?

lib/src/query_service.dart
ruben-aghayan: bump

test/insertion_order_parser_test.dart
```diff
@@ -0,0 +1,293 @@
...
+
+      test('result is an insertionOrder with only advertiserId', () {
+        final actual = InsertionOrderParser.parse(input);
+
+        final expected = insertionOrderTemplate..advertiserId = '111111';
+        expect(actual, expected);
+      });
+    });
+
+    group('parse insertionOrder from json that contains everything:', () {
```
ruben-aghayan: group are used to share setup/teardown between multiple tests and logically group them.
if you have groups of 1 test, consider putting in all of the logic inside the 1 test in a traditional arrange,act, assert setup.

e.g. so if you had another test that verified something about the complete input then it would make sense to share the setup

lib/src/query_service.dart
```diff
@@ -0,0 +1,36 @@
...
+
+    await GoogleAPI.client.request(requestArgs).then(allowInterop((response) {
+      final rawResponse = stringify(response.result);
+      _insertionOrderLists.add(InsertionOrderParser.parse(rawResponse));
+      print(_insertionOrderLists[0]);
+    }));
+  }
+
+  /// Generates query based on user inputs.
+  String _generateQuery(String advertiserId, String insertionOrderId) {
```
ruben-aghayan: can this be a static method?

lib/src/query_service.dart
```diff
@@ -0,0 +1,36 @@
...
+import 'json_js.dart';
+import 'insertion_order_parser.dart';
+import 'proto/insertion_order.pb.dart';
+
+class QueryService {
+  QueryService._private();
+
+  static final QueryService _singleton = QueryService._private();
+
+  factory QueryService() {
```
ruben-aghayan: prefer using @Injectable rather than using singleton pattern
same in a couple other places.

ruben-aghayan (changes requested):

lib/src/insertion_order_parser.dart
hanj7221: ugh now this is tested

lib/src/query_service.dart
hanj7221: New issue created #31 and will be resolved in a separate PR.

lib/src/query_service.dart
```diff
@@ -0,0 +1,36 @@
...
+
+  /// Executes the query and passes the returned result to the parser.
+  void execQuery(String advertiserId, String insertionOrderId) async {
+    final requestArgs = RequestArgs(
+        path: _generateQuery(advertiserId, insertionOrderId), method: 'GET');
+
+    await GoogleAPI.client.request(requestArgs).then(allowInterop((response) {
+      final rawResponse = stringify(response.result);
+      _insertionOrderLists.add(InsertionOrderParser.parse(rawResponse));
+      print(_insertionOrderLists[0]);
```
ruben-aghayan: rm debug code

lib/src/parser.dart
hanj7221: New issue created #30, will be resolved in a separate PR

lib/src/query_service.dart
hanj7221: Singleton has been removed for this class, and will do the same for other places in a separate PR (new issue #32)

lib/src/insertion_order_parser.dart
```diff
@@ -0,0 +1,164 @@
...
+    final date =
+        InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment_DateRange_Date()
+          ..year = map['year'] ?? _emptyEntry
+          ..month = map['month'] ?? _emptyEntry
+          ..day = map['day'] ?? _emptyEntry;
+
+    return date;
+  }
+
+  static InsertionOrder_EntityStatus _createEntityStatus(String target) {
```
ruben-aghayan: consider: rather than having the one line method just doing it inline. not required though.

also: is there a default value method on proto enums?

hanj7221: I don't want to believe this either but I didn't find a default value method :(

Instead of enum.values[0], would enum.valueOf(0) sounds more like a default value?

lib/src/credential_component.dart
millerju-google: Why don't we want to use material-button? It's the button component we use over the much older <button> component.

lib/src/parser.dart
millerju-google: I agree, something like Dv360ServiceResponseParser?

lib/src/parser.dart
millerju-google: Rather than throwing, I generally prefer 'reasonable defaults'. For something like this it doesn't have to be 'all or nothing'.
You can fill in the IO like this:
  final insertionOrder = InsertionOrder()
      ..advertiserId = map['advertiserId'] ?? 'Unknown'
      ..campaignId = map['campaignId'] ?? 'Unknown'
     .... and so on

lib/src/credential_component.dart
hanj7221: The reason for this change is documented here #23. Using button is a temporary solution for now so that we can compile, we are still looking for better options though.

lib/src/parser.dart
millerju-google: Is the type in map['budgetSegments'], is a list? If so you can use the map() method:
  budget.budgetSegments = map['budgetSegments'].map(createBudgetSegment).toList();

I think however that the variable name 'map' is pretty confusing, you can instead call it something like jsonData or something like that instead.

lib/src/parser.dart
millerju-google: I also like the idea of putting in default values if we don't know them (though budget segements should always have dates):
return InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment_DateRange()
  ..startDate = createDate(map['startDate'])
  ..endDate = createDate(map['endDate']);
.....
  createDate handles null input by returning a default date value (e.g. Epoch 0 time).

lib/src/parser.dart
millerju-google: In this case if any of the values are missing, the entire value is invalid and we should return a default date:
   if (map == null || !map.containsKey('year') || !map.containsKey('month') .. {
     // empty proto
     return InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment_DateRange_Date();
  }
 return InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment_DateRange_Date()
     ..year = map['year']
     ....;

lib/src/parser.dart
```diff
@@ -0,0 +1,222 @@
...
+    for (var status in InsertionOrder_EntityStatus.values) {
+      if (status.name == target) {
+        return status;
+      }
+    }
+    throw FormatException();
```
millerju-google: I think a better value is to return the 'unknown' or 'unspecified' default value, thats what our proto parses do in this case. This can happen as enums evolve over time and there is a mismatch between the client library and the server library.

lib/src/insertion_order_parser.dart
millerju-google: I think it's better to ensure the map is actually typed:
  static final PACING_PERIOD_MAP = Map<String, InsertionOrder_Pacing_PacingPeriod> 
     .fromIterable(InsertionOrder_Pacing_PacingPeriod.values, k -> k.name, v -> v);
If you want to stick with your current syntax, make sure to use 'final' instead of 'var' in the for loops.

lib/src/insertion_order_parser.dart
millerju-google: By convention the '0' value is usally the default, however I think it's better to be explicit here as not all of these enums are doing this, so something like this instead:
   return ENTITY_STATUS_MAP[target] ?? InsertionOrder_EntityStatus.UNKOWN

test/insertion_order_parser_test.dart
```diff
@@ -0,0 +1,302 @@
...
+      final advertiserId = '"advertiserId":"11111"';
+      final campaignId = '"campaignId":"2222222"';
+      final insertionOrderId = '"insertionOrderId":"3333333"';
+      final displayName = '"displayName":"display name"';
+      final entityStatus = '"entityStatus":"ENTITY_STATUS_ACTIVE"';
+      final updateTime = '"updateTime":"2020-06-23T17:14:58.685Z"';
+      final pacing = '"pacing":{"pacingPeriod":"PACING_PERIOD_FLIGHT",'
+          '"pacingType":"PACING_TYPE_AHEAD"}';
+      final budgetSegment = '"budgetSegments":['
+          '{"budgetAmountMicros":"4000000","description":"year-2019",'
```
millerju-google: For multiline string constants you can use the triple quotes like this:
final budgetSegment = """
   "budgetSegments": [
       {"budgetAmountMicros":....
""";

test/insertion_order_parser_test.dart
```diff
@@ -0,0 +1,302 @@
...
+          ..pacingPeriod =
+              InsertionOrder_Pacing_PacingPeriod.PACING_PERIOD_FLIGHT
+          ..pacingType = InsertionOrder_Pacing_PacingType.PACING_TYPE_AHEAD
+          ..dailyMaxImpressions = emptyEntry;
+        final expected = insertionOrderTemplate..pacing = pacing;
+        expect(actual, expected);
+      });
+
+      test('pacingPeriod, pacingType and dailyMaxMicros', () {
+        input = '{"pacing":{"pacingPeriod":"PACING_PERIOD_FLIGHT",'
```
millerju-google: Are you sure this is what the service returns? When we make API calls from the DV360 Web UI the protos are all converted to 'lite' json which uses numbers to return the enum values instead of strings.

millerju-google (approved):
Looks good, I just added a couple of optional comments.

lib/src/insertion_order_parser.dart
```diff
@@ -0,0 +1,164 @@
...
+    return _budgetUnitMap[target] ??
+        InsertionOrder_InsertionOrderBudget_BudgetUnit.values[0];
+  }
+
+  static InsertionOrder_InsertionOrderBudget_InsertionOrderAutomationType
+      _createAutomationType(String target) {
+    /// Returns default value if [target] is null
+    if (target == null) {
+      return InsertionOrder_InsertionOrderBudget_InsertionOrderAutomationType
+          .values[0];
```
arulprakash-g: Looks like default should be INSERTION_ORDER_AUTOMATION_TYPE_NONE

arulprakash-g (approved):

lib/src/parser.dart
hanj7221: Yeah, that var name 'map' is confusing and has now been changed.  Unfortunately map() cannot be used for this field, since it is denoted as repeated in the proto definition, and Dart doesn't generate setters for repeated fields. It can only call add(). But I will keep in mind to take advantage of this map() method in other cases that are applicable!

lib/src/insertion_order_parser.dart
hanj7221: Yes, automationType.values[0] here evaluates to INSERTION_ORDER_AUTOMATION_TYPE_NONE

test/insertion_order_parser_test.dart
hanj7221: Yes, that's what get returned when I use the google javascript client library to make those API calls. I can double check this with the API team.

ruben-aghayan (approved):
lgtm

PR #33, opened 2020-06-30
Extend excel class to do population

### LINKED ISSUE #6: Extends service class to handle excel population for one IO [1~2 day]
Extends the existing query submission service class so that it can use the returned results to populate the spreadsheet accordingly.

hanj7221:
Closing #6

New issues that arose in this PR and will be resolved in a separate PR:

- Conditional format for the budget amount column: #34 
- Only sum up active budget segments: #35 
- Name for new sheet needs to be unique: #36


lib/src/excel.dart
```diff
@@ -1,36 +1,128 @@
...
 class ExcelDart {
-  ExcelDart._private();
+  static final _startAddressInt = 'A'.codeUnitAt(0);
+  static final _fontName = 'Roboto';
+  static final _fontSize = 12;
+  static final _horizontalAlignment = 'Center';
+  static final _borderStyle = 'Continuous';
 
-  static final ExcelDart _singleton = ExcelDart._private();
+  static final _tableHeader = [
```
ruben-aghayan, line 25: does this have to keep in line with the row proto? what happens if they mismatch?

lib/src/excel.dart
```diff
@@ -1,36 +1,128 @@
...
 
-  factory ExcelDart() {
-    return _singleton;
-  }
-
-  /// Waits for Office APIs to be ready and then executes function [_populate].
-  void exec() async {
+  /// Waits for Office APIs to be ready and then populates the spreadsheet with
+  /// entries in the [insertionOrderList]
+  static void populate(List insertionOrderList) async {
```
ruben-aghayan: add type parameter

lib/src/excel.dart
```diff
@@ -1,36 +1,128 @@
...
     await Office.onReady(allowInterop((info) async {
-      await ExcelJS.run(allowInterop(_populate));
+      await ExcelJS.run(allowInterop((context) {
+        // Grabs the current worksheet
```
ruben-aghayan: nit: comment should be formatted as sentences

lib/src/excel.dart
```diff
@@ -1,36 +1,128 @@
...
+        final sheet = context.workbook.worksheets.getActiveWorksheet();
```
ruben-aghayan: consider: making a new sheet to avoid problems of overwriting/conflicting with existing data

lib/src/excel.dart
```diff
@@ -1,36 +1,128 @@
...
+          ..format.borders.getItem('EdgeBottom').style = _borderStyle;
+
+        // Adds and formats table body
+        table.getDataBodyRange()
+          ..formulas = _generateTableBody(insertionOrderList)
+          ..format.font.name = _fontName
+          ..format.font.size = _fontSize
+          ..format.horizontalAlignment = _horizontalAlignment;
+
+        // Auto-fits all used cells
```
ruben-aghayan: love the comments especially as a reviewer not familiar with excel api

lib/src/excel.dart
```diff
@@ -1,36 +1,128 @@
...
-    final values = <List<String>>[
-      ['test test']
-    ];
-    range.values = values;
-    range.format.autofitColumns();
-    return context.sync();
+  static List _generateTableBody(List insertionOrderList) {
+    final tableBody = [];
+
+    for (InsertionOrder io in insertionOrderList) {
```
ruben-aghayan: what about doing something like insertionOrderList.map(conversionfunction).toList() do you think that would be more idiomatic?

lib/src/excel.dart
```diff
@@ -1,36 +1,128 @@
...
+        _getEndDate(io.budget.budgetSegments.last),
+      ]);
+    }
+    return tableBody;
+  }
+
+  static String _calculateTotalBudgetAmount(
+      List<InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment>
+          budgetSegments) {
+    var totalBudgetMicros = Int64.ZERO;
```
ruben-aghayan: nice use of the constant

lib/src/excel.dart
```diff
@@ -1,36 +1,128 @@
...
+
+    for (var segment in budgetSegments) {
```
ruben-aghayan: what about using a reducer?

lib/src/excel.dart
```diff
@@ -1,36 +1,128 @@
...
+      totalBudgetMicros += Int64.parseInt(segment.budgetAmountMicros);
+    }
+
+    return (totalBudgetMicros.toDouble() / 1000000).toString();
```
ruben-aghayan: is the toDouble necessary for division? that could result in some precision loss
consider storing the 1000000 as a constant. or having a conversion function
also does this format it like we would expect money to be formatted? ie $##.## ?

ruben-aghayan (changes requested):
since we dont have a testing story for the pieces interacting with excel, consider writing (brief) instructions for manual testing with some expectations

lib/src/excel.dart
arulprakash-g: Also currency vs impression budgets?

lib/src/excel.dart
```diff
@@ -1,36 +1,128 @@
...
+        io.budget.automationType.toString(),
+        _calculateTotalBudgetAmount(io.budget.budgetSegments),
+        _getStartDate(io.budget.budgetSegments.first),
+        _getEndDate(io.budget.budgetSegments.last),
+      ]);
+    }
+    return tableBody;
+  }
+
+  static String _calculateTotalBudgetAmount(
```
arulprakash-g, line 115: I think this is not very relevant. Advertisers are mostly interested in their currently active segment.

lib/src/excel.dart
arulprakash-g: Also impression vs currency budget.

lib/src/excel.dart
```diff
@@ -1,36 +1,128 @@
...
+      tableBody.add([
+        io.insertionOrderId,
+        io.advertiserId,
+        io.campaignId,
+        io.displayName,
+        io.entityStatus.toString(),
+        io.updateTime,
+        io.pacing.pacingPeriod.toString(),
+        io.pacing.pacingType.toString(),
+        io.pacing.dailyMaxMicros,
```
arulprakash-g: This is in micros and total budget we divide by 10^6 - lets use a consistent unit.

arulprakash-g (approved):

lib/src/excel.dart
hanj7221: The formatting will be done from the Excel side using their javascript APIs. Conditional format the budget amount column based on the budgetType (currency or impression) will be done in a separate PR. New issue created #34.

lib/src/excel.dart
hanj7221: tableHeader will need to be in line with the _generateTableRow() function. Changing the protobuff definitions can affect the _generateTableRow() function (invalid getters etc). I will add a comment specifying this.

lib/src/excel.dart
hanj7221: New issue created #35, will be addressed in a separate PR.

lib/src/excel.dart
```diff
@@ -33,20 +40,26 @@ class ExcelDart {
...
+  /// a new spreadsheet with [sheetName] and populates the spreadsheet with
   /// entries in the [insertionOrderList]
-  static void populate(List insertionOrderList) async {
+  static void populate(List<InsertionOrder> insertionOrderList) async {
     await Office.onReady(allowInterop((info) async {
       await ExcelJS.run(allowInterop((context) {
-        // Grabs the current worksheet
-        final sheet = context.workbook.worksheets.getActiveWorksheet();
+        // Adds a new worksheet
+        final sheet = context.workbook.worksheets.add('Query');
```
hanj7221, line 50: Each new sheet needs a unique name. New issue created for this #36, and will be addressed in a separate PR.

lib/src/excel.dart
ruben-aghayan: i think we want to handle currency budget case end to end first then impression budget?

lib/src/excel.dart
ruben-aghayan: ok lets make sure #35 doesnt fall off since its part of our mvp

lib/src/excel.dart
```diff
@@ -1,36 +1,139 @@
...
+  static final _startAddressInt = 'A'.codeUnitAt(0);
+  static final _fontName = 'Roboto';
+  static final _fontSize = 12;
+  static final _horizontalAlignment = 'Center';
+  static final _borderStyle = 'Continuous';
 
-  static final ExcelDart _singleton = ExcelDart._private();
+  /// Header for the master table.
+  ///
+  /// If changes have been to [insertion_order.proto] or
```
ruben-aghayan: have been made to

lib/src/excel.dart
```diff
@@ -1,36 +1,139 @@
...
-  Future<void> _populate(RequestContext context) async {
-    final sheet = context.workbook.worksheets.getActiveWorksheet();
-    final range = sheet.getRange('A1');
-    final values = <List<String>>[
-      ['test test']
-    ];
-    range.values = values;
-    range.format.autofitColumns();
-    return context.sync();
+  static List _generateTableRow(InsertionOrder io) => [
```
ruben-aghayan: add type parameter

lib/src/excel.dart
```diff
@@ -1,36 +1,139 @@
...
+
+  static String _calculatePacingDailyMax(String dailyMaxMicros) =>
+      dailyMaxMicros.isEmpty
+          ? dailyMaxMicros
+          : (Int64.parseInt(dailyMaxMicros) * 1e-6).toString();
+
+  static String _calculateTotalBudgetAmount(
+      List<InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment>
+          budgetSegments) {
+    var budgetAmountList = budgetSegments
```
ruben-aghayan: can this be final?

lib/src/excel.dart
```diff
@@ -1,36 +1,139 @@
...
+        .map((segment) => Int64.parseInt(segment.budgetAmountMicros))
+        .toList();
+    var totalBudgetMicros =
```
ruben-aghayan: same

lib/src/excel.dart
ruben-aghayan: you can drop a todo with the issue on the method

lib/src/excel.dart
```diff
@@ -1,36 +1,139 @@
...
+        budgetAmountList.reduce((value, element) => value + element);
+
+    return (totalBudgetMicros.toDouble() * 1e-6).toString();
```
ruben-aghayan, line 124: same with leaving a todo to handle the formatting

ruben-aghayan (approved):

lib/src/excel.dart
ruben-aghayan: can you merge these list comps?
reduce (value + Int64(parseInt(element.budgetAmountMicros))

hanj7221: Reducer needs to return the same type as the list, but now the map & reduce are chained together so we don't the intermediate variable.

Issue #36, opened 2020-07-01
Excel spreadsheet needs to be created with unique names


hanj7221:
Need to way to make sure that, for each query, a unique name is given to the new sheet/table getting created.

PR #37, opened 2020-07-02
Adding functions to calculate active budget segments and currency formats

### LINKED ISSUE #34: Add conditional format to budget amount columns based on budget type
Use Excel javascript apis to do conditional formatting for budget amount columns.

### LINKED ISSUE #35: Change calculateTotalBudgetAmount function to only consider active budget segments
Active budget segments who has start date < now < end date.

hanj7221:
Closes #35 and #34


lib/src/excel.dart
```diff
@@ -45,7 +46,9 @@ class ExcelDart {
   /// entries in the [insertionOrderList]
   static void populate(List<InsertionOrder> insertionOrderList) async {
     await Office.onReady(allowInterop((info) async {
-      await ExcelJS.run(allowInterop((context) {
+      final tableName = 'Performance_Data_Table';
```
hanj7221, line 52: Table name also needs to be distinct, adding this to #36. Maybe new sheet and new table can share a same name, since they both need to be distinct?

lib/src/excel.dart
```diff
@@ -76,66 +80,98 @@ class ExcelDart {
...
       }));
+
+      // Conditionally format the budget column once the table is set.
+      _formatBudgetColumn(tableName);
+    }));
+  }
+
+  /// Conditionally formats the Budget column based on Budget Unit.
+  ///
+  /// RC[-2] in the formula follows Excel's Relative Notation and references
```
hanj7221: Should be "RC[-2]" in quotes so it doesn't confuse with Dart doc style.

lib/src/query_service.dart
```diff
@@ -1,6 +1,7 @@
 import 'dart:js';
 
 import 'package:angular/angular.dart';
+import 'package:dv360_excel_plugin/src/excel.dart';
```
hanj7221: Should be relative import

lib/src/excel.dart
```diff
@@ -76,66 +80,98 @@ class ExcelDart {
           ..format.font.size = _fontSize
           ..format.horizontalAlignment = _horizontalAlignment;
 
+        // Formats the Daily Max column as currency.
+        table.columns.getItem('Daily Max').getRange().numberFormat = _currencyFormat;
```
ruben-aghayan: nice

lib/src/excel.dart
```diff
@@ -76,66 +80,98 @@ class ExcelDart {
...
+      // Conditionally format the budget column once the table is set.
+      _formatBudgetColumn(tableName);
+    }));
+  }
+
+  /// Conditionally formats the Budget column based on Budget Unit.
+  ///
+  /// RC[-2] in the formula follows Excel's Relative Notation and references
+  /// the cell that is in the same row but its column number shifted left by 2.
+  /// Here the offset between the Budget column and Budget Type Column is -2.
```
ruben-aghayan, line 108: nice comment

lib/src/excel.dart
```diff
@@ -15,6 +15,7 @@ class ExcelDart {
   static final _fontSize = 12;
```
ruben-aghayan: these could be const right?

lib/src/excel.dart
```diff
@@ -76,66 +80,98 @@ class ExcelDart {
...
+      insertionOrder.displayName,
+      insertionOrder.entityStatus.toString(),
+      insertionOrder.updateTime,
+      insertionOrder.pacing.pacingPeriod.toString(),
+      insertionOrder.pacing.pacingType.toString(),
+      _calculatePacingDailyMax(insertionOrder.pacing.dailyMaxMicros),
+      insertionOrder.pacing.dailyMaxImpressions
+    ];
+
+    row.addAll(_generateBudgetColumns(insertionOrder.budget));
```
ruben-aghayan: this could be an opportunity to use the spread operator

lib/src/excel.dart
```diff
@@ -76,66 +80,98 @@ class ExcelDart {
...
 
-  /// Todo: calculate budget total for active segments only.
-  /// Issue: https://github.com/googleinterns/dv360-excel-plugin/issues/35.
-  static String _calculateTotalBudgetAmount(
-      List<InsertionOrder_InsertionOrderBudget_InsertionOrderBudgetSegment>
-          budgetSegments) {
-    final totalBudgetMicros = budgetSegments
+  /// Generates columns for budget unit, automation type,
+  /// active segments total budget, start date and end date from [budget]
+  static List<String> _generateBudgetColumns(
```
ruben-aghayan: why combine these all into one method vs having them split out as before?

lib/src/query_service.dart
```diff
@@ -17,8 +18,9 @@ class QueryService {
         path: _generateQuery(advertiserId, insertionOrderId), method: 'GET');
 
     await GoogleAPI.client.request(requestArgs).then(allowInterop((response) {
-      final rawResponse = stringify(response.result);
-      _insertionOrderLists.add(InsertionOrderParser.parse(rawResponse));
+      _insertionOrderLists
+          .add(InsertionOrderParser.parse(stringify(response.result)));
+      ExcelDart.populate(_insertionOrderLists);
```
ruben-aghayan: i dont think service should populate. whatever called the service should populate and parse. this uncouples the service from the excel piece.

ruben-aghayan (changes requested):

lib/src/excel.dart
hanj7221: Hmmm...nothing really, but now I look at it and agree that splitting them out is more readable.

lib/src/excel.dart
```diff
@@ -146,32 +153,39 @@ class ExcelDart {
...
+
+  static String _calculateTotalBudgetAmount(
+      List<InsertionOrder_Budget_BudgetSegment> activeSegments) {
     final totalBudgetMicros = activeSegments
         .map((segment) => Int64.parseInt(segment.budgetAmountMicros))
         .reduce((sumAmount, segmentAmount) => sumAmount + segmentAmount);
 
-    final startDate = activeSegments.first.dateRange.startDate;
-    final endDate = activeSegments.last.dateRange.endDate;
+    return (totalBudgetMicros.toDouble() * 1e-6).toString();
```
ruben-aghayan: is there accuracy loss from this? reporting api will also return in micros. so we should make converting micros into a util function and handle it carefully in one place

lib/src/query_component.dart
```diff
@@ -19,12 +22,17 @@ import 'query_service.dart';
 )
 class QueryComponent {
   final buttonName = 'populate';
-  final QueryService queryService;
+  final QueryService _queryService;
```
ruben-aghayan, line 25: nice

lib/src/query_service.dart
```diff
@@ -1,27 +1,25 @@
...
-import 'proto/insertion_order.pb.dart';
 
 @Injectable()
 class QueryService {
-  static final _insertionOrderLists = <InsertionOrder>[];
-
-  /// Executes the query and passes the returned result to the parser.
-  void execQuery(String advertiserId, String insertionOrderId) async {
+  /// Executes the query and returns the raw javascript object.
+  Future<Object> execQuery(String advertiserId, String insertionOrderId) async {
```
ruben-aghayan: why object vs dynamic?

ruben-aghayan: can you make exec query a generic so the return type is stricter?

lib/src/query_component.dart
```diff
@@ -19,12 +22,17 @@ import 'query_service.dart';
...
 
-  QueryComponent(this.queryService);
+  QueryComponent(this._queryService);
 
-  void onClick() => queryService.execQuery(advertiserId, insertionOrderId);
+  void onClick() async {
+    final jsonResponse =
+        await _queryService.execQuery(advertiserId, insertionOrderId);
+    final parsedResponse = InsertionOrderParser.parse(stringify(jsonResponse));
+    ExcelDart.populate([parsedResponse]);
```
ruben-aghayan: does this return a future?

ruben-aghayan (changes requested):

lib/src/excel.dart
```diff
@@ -84,57 +91,99 @@ class ExcelDart {
...
 
+  /// Calculates pacing daily max in standard unit.
   static String _calculatePacingDailyMax(String dailyMaxMicros) =>
       dailyMaxMicros.isEmpty
           ? dailyMaxMicros
           : (Int64.parseInt(dailyMaxMicros) * 1e-6).toString();
 
-  /// Todo: calculate budget total for active segments only.
-  /// Issue: https://github.com/googleinterns/dv360-excel-plugin/issues/35.
+  /// Extracts active budget segments based on start and end dates.
```
arulprakash-g: There can only be 1 active budget segment at a time.

lib/src/excel.dart
```diff
@@ -84,57 +91,99 @@ class ExcelDart {
...
+  static List<InsertionOrder_Budget_BudgetSegment> _extractActiveBudgetSegments(
+      List<InsertionOrder_Budget_BudgetSegment> budgetSegments) {
```
arulprakash-g: Related to above - don't return list.

lib/src/excel.dart
arulprakash-g: Can you check how impression budget IOs are represented (ie are they also in micros?)

arulprakash-g (approved):

lib/src/excel.dart
hanj7221: Yes, seems like impression based budgets are also in micros per the [documentation](https://developers.google.com/display-video/api/reference/rest/v1/advertisers.insertionOrders#InsertionOrderBudgetSegment).

hanj7221: That .toDouble() shouldn't be there. Sorry!

hanj7221: That what it says in the documentation: On some platforms, inputs with large absolute values (i.e., > 2^52) may lose some of their low-order bits.

hanj7221: Please see the latest commit for how this is changed.

lib/src/excel.dart
```diff
@@ -84,59 +95,93 @@ class ExcelDart {
...
-          budgetSegments) {
-    final totalBudgetMicros = budgetSegments
-        .map((segment) => Int64.parseInt(segment.budgetAmountMicros))
-        .reduce((value, element) => value + element);
-
-    /// Todo: conditional formatting for currency.
-    /// Issue: https://github.com/googleinterns/dv360-excel-plugin/issues/34.
-    return (totalBudgetMicros.toDouble() * 1e-6).toString();
-  }
+          : Util.convertMicrosToStandardUnitString(
```
ruben-aghayan, line 155: nice

lib/src/query_component.dart
```diff
@@ -14,17 +17,23 @@ import 'query_service.dart';
...
   String advertiserId;
   String insertionOrderId;
 
-  QueryComponent(this.queryService);
+  QueryComponent(this._queryService, this._excel);
 
-  void onClick() => queryService.execQuery(advertiserId, insertionOrderId);
+  void onClick() async {
+    final jsonResponse =
+        await _queryService.execQuery(advertiserId, insertionOrderId);
```
ruben-aghayan, line 35: add type parameter to generic method e.g. execQuery<String>(advertiserId)

ruben-aghayan (approved):

PR #47, opened 2020-07-07
Create proto for reporting query params

hanj7221:


ruben-aghayan (changes requested):
can this go in with your other protos?

ruben-aghayan (approved):

PR #50, opened 2020-07-09
Query, parse and populate spent data using reporting api

### LINKED ISSUE #7: Query the spending data using reporting API [1~3 day]
Cost estimate: 1~3 day

hanj7221:
Closes #7

### DIFF
--- a/lib/src/credential_service.dart
+++ b/lib/src/credential_service.dart
@@ -10,16 +10,16 @@ class CredentialService {
   static GoogleAuth _googleAuth;
   static const _redirectURI = 'http://localhost:8080';
 
-  static final _dv3Scope = 'https://www.googleapis.com/auth/display-video';
-  static final _dbmScope =
+  static const _dv3Scope = 'https://www.googleapis.com/auth/display-video';
+  static const _dbmScope =
       'https://www.googleapis.com/auth/doubleclickbidmanager';
-  static final _googleStorageScope =
+  static const _googleStorageScope =
       'https://www.googleapis.com/auth/devstorage.read_only';
-  static final _scope = '$_dv3Scope $_dbmScope $_googleStorageScope';
+  static const _scope = '$_dv3Scope $_dbmScope $_googleStorageScope';
 
-  static final _dv3DiscoveryUrl =
+  static const _dv3DiscoveryUrl =
       'https://displayvideo.googleapis.com/\$discovery/rest?version=v1';
-  static final _dbmDiscoveryUrl =
+  static const _dbmDiscoveryUrl =
       'https://content.googleapis.com/discovery/v1/apis/doubleclickbidmanager/v1.1/rest';
 
   /// Loads gapi.client library and calls [_initClient]
--- a/lib/src/credential_service.dart
+++ b/lib/src/credential_service.dart
@@ -8,8 +8,19 @@ import 'private_keys.dart' as key_store;
 @Injectable()
 class CredentialService {
   static GoogleAuth _googleAuth;
-  static const _scope = 'https://www.googleapis.com/auth/display-video';
   static const _redirectURI = 'http://localhost:8080';
+
+  static final _dv3Scope = 'https://www.googleapis.com/auth/display-video';
+  static final _dbmScope =
+      'https://www.googleapis.com/auth/doubleclickbidmanager';
+  static final _googleStorageScope =
+      'https://www.googleapis.com/auth/devstorage.read_only';
+  static final _scope = '$_dv3Scope $_dbmScope $_googleStorageScope';
+
+  static final _dv3DiscoveryUrl =
+      'https://displayvideo.googleapis.com/\$discovery/rest?version=v1';
+  static final _dbmDiscoveryUrl =
+      'https://content.googleapis.com/discovery/v1/apis/doubleclickbidmanager/v1.1/rest';
 
   /// Loads gapi.client library and calls [_initClient]
   /// when library finishes loading.
@@ -29,17 +40,13 @@ class CredentialService {
   }
 
   static void _initClient() async {
-    // Retrieve the discovery document for version 1 of DV360 public API.
-    final discoveryUrl =
-        'https://displayvideo.googleapis.com/\$discovery/rest?version=v1';
-
     // Initializes the gapi.client object, which app uses to make API requests.
     // [apikey] and [clientId] are obtained from google api console.
     // [scope] specifies space-delimited list of access scopes.
     final initArgs = InitArgs(
         apiKey: key_store.apiKey,
         clientId: key_store.clientID,
-        discoveryDocs: [discoveryUrl],
+        discoveryDocs: [_dv3DiscoveryUrl, _dbmDiscoveryUrl],
         scope: _scope);
 
     await GoogleAPI.client.init(initArgs).then(allowInterop((value) {
--- a/lib/src/excel.dart
+++ b/lib/src/excel.dart
@@ -40,6 +40,7 @@ class ExcelDart {
     'Budget Unit',
     'Automation Type',
     'Budget',
+    'Spent',
     'Start Date',
     'End Date'
   ];
@@ -83,8 +84,10 @@ class ExcelDart {
           ..format.font.size = _fontSize
           ..format.horizontalAlignment = _horizontalAlignment;
 
-        // Formats the Daily Max column as currency.
+        // Formats the Daily Max and Spent column as currency.
         table.columns.getItem('Daily Max').getRange().numberFormat =
+            _currencyFormat;
+        table.columns.getItem('Spent').getRange().numberFormat =
             _currencyFormat;
 
         // Auto-fits all used cells.
@@ -126,8 +129,7 @@ class ExcelDart {
   /// Generates a row from [insertionOrder] by creating fields that matches
   /// those specified in [_tableHeader].
   static List<String> _generateTableRow(InsertionOrder insertionOrder) {
-    final activeBudgetSegment =
-        _extractActiveBudgetSegment(insertionOrder.budget.budgetSegments);
+    final activeBudgetSegment = insertionOrder.budget.budgetSegments.first;
 
     return [
       insertionOrder.insertionOrderId,
@@ -143,6 +145,7 @@ class ExcelDart {
       insertionOrder.budget.budgetUnit.toString(),
       insertionOrder.budget.automationType.toString(),
       _calculateActiveBudgetAmount(activeBudgetSegment.budgetAmountMicros),
+      insertionOrder.spent,
       _calculateDate(activeBudgetSegment.dateRange.startDate),
       _calculateDate(activeBudgetSegment.dateRange.endDate),
     ];
@@ -154,23 +157,6 @@ class ExcelDart {
           ? dailyMaxMicros
           : Util.convertMicrosToStandardUnitString(
               Int64.parseInt(dailyMaxMicros));
-
-  /// Extracts the active budget segment based on start and end dates.
-  ///
-  /// There can only be 1 active budget segment at a time.
-  static InsertionOrder_Budget_BudgetSegment _extractActiveBudgetSegment(
-      List<InsertionOrder_Budget_BudgetSegment> budgetSegments) {
-    final now = DateTime.now();
-    final activeSegments = budgetSegments.where((segment) {
-      final startDate = segment.dateRange.startDate;
-      final endDate = segment.dateRange.endDate;
-      return now.isAfter(
-              DateTime.utc(startDate.year, startDate.month, startDate.day)) &&
-          now.isBefore(DateTime.utc(endDate.year, endDate.month, endDate.day));
-    });
-
-    return activeSegments.first;
-  }
 
   static String _calculateActiveBudgetAmount(String budgetAmountMicros) =>
       Util.convertMicrosToStandardUnitString(
--- a/lib/src/excel.dart
+++ b/lib/src/excel.dart
@@ -129,8 +129,6 @@ class ExcelDart {
   /// Generates a row from [insertionOrder] by creating fields that matches
   /// those specified in [_tableHeader].
   static List<String> _generateTableRow(InsertionOrder insertionOrder) {
-    final activeBudgetSegment = insertionOrder.budget.budgetSegments.first;
-
     return [
       insertionOrder.insertionOrderId,
       insertionOrder.advertiserId,
@@ -144,10 +142,13 @@ class ExcelDart {
       insertionOrder.pacing.dailyMaxImpressions,
       insertionOrder.budget.budgetUnit.toString(),
       insertionOrder.budget.automationType.toString(),
-      _calculateActiveBudgetAmount(activeBudgetSegment.budgetAmountMicros),
+      _calculateActiveBudgetAmount(
+          insertionOrder.budget.activeBudgetSegment.budgetAmountMicros),
       insertionOrder.spent,
-      _calculateDate(activeBudgetSegment.dateRange.startDate),
-      _calculateDate(activeBudgetSegment.dateRange.endDate),
+      _calculateDate(
+          insertionOrder.budget.activeBudgetSegment.dateRange.startDate),
+      _calculateDate(
+          insertionOrder.budget.activeBudgetSegment.dateRange.endDate),
     ];
   }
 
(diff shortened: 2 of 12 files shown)


lib/src/credential_service.dart
```diff
@@ -8,9 +8,20 @@ import 'private_keys.dart' as key_store;
 @Injectable()
 class CredentialService {
   static GoogleAuth _googleAuth;
-  static const _scope = 'https://www.googleapis.com/auth/display-video';
   static const _redirectURI = 'http://localhost:8080';
 
+  static final _dv3Scope = 'https://www.googleapis.com/auth/display-video';
```
hanj7221: These can all be static const

lib/src/insertion_order_parser.dart
```diff
@@ -82,8 +82,18 @@ class InsertionOrderParser {
       ..budgetUnit = _createBudgetUnit(map['budgetUnit'])
       ..automationType = _createAutomationType(map['automationType']);
 
+    final now = DateTime.now();
     for (Map segmentMap in map['budgetSegments'] ?? []) {
-      budget.budgetSegments.add(_createBudgetSegment(segmentMap));
+      final segment = _createBudgetSegment(segmentMap);
```
hanj7221: Moving this active segment calculation from population to parsing. Shall we modify the proto definition to just make budgetSegment optional instead of repeated since there is only gonna be one active segment at a time?

lib/src/reporting_query_parser.dart
```diff
@@ -0,0 +1,42 @@
...
+
+  /// Parse google storage download from a json string.
+  static String parseDownloadPathFromJsonString(String jsonString) =>
+      (json.decode(jsonString)['metadata'] ??
+          const {})['googleCloudStoragePathForLatestReport'] ??
+      _emptyEntry;
+
+  /// Parse revenue from a json string.
+  static Map<String, String> parseRevenueFromString(String responseFull) {
+    String responseBody =
```
hanj7221: Let me know if this is readable? The intent is do a nested null operator on Maps :(

lib/src/reporting_query_parser.dart
```diff
@@ -0,0 +1,42 @@
...
+        ((json.decode(responseFull)['gapiRequest'] ?? const {})['data'] ??
+                const {})['body'] ??
+            '';
+
+    if (responseBody.isEmpty) return <String, String>{};
+
+    // Extracts the substring after 'Revenue (USD)' and before 'Report Time',
```
hanj7221: Here is a sample output:

Insertion Order ID,Revenue (USD)
1234567,100.00

Report Time:,2020/07/09 01:14 GMT
Date Range:,2020/06/22 to 2020/07/21
Group By:,Insertion Order ID
MRC Accredited Metrics,Active View metrics are accredited only when Measurement Source = Measured
"Reporting numbers from the previous month are finalized (for billing purposes) by the third of each month, unless communicated otherwise."
Filter by Advertiser ID:,164337
Filter by Insertion Order ID:,8127549

lib/src/reporting_query_parser.dart
```diff
@@ -0,0 +1,42 @@
+import 'dart:convert';
```
hanj7221, line 1: Unit test for this file will be added once I get back some reviews about the way I currently use to parse revenue from response. I don't like how I currently handle it, and I'm looking forward to any suggestions.

lib/src/insertion_order_parser.dart
ruben-aghayan: yea +1
feel free to change in this pr if you want

lib/src/excel.dart
```diff
@@ -126,8 +129,7 @@ class ExcelDart {
   /// Generates a row from [insertionOrder] by creating fields that matches
   /// those specified in [_tableHeader].
   static List<String> _generateTableRow(InsertionOrder insertionOrder) {
-    final activeBudgetSegment =
-        _extractActiveBudgetSegment(insertionOrder.budget.budgetSegments);
+    final activeBudgetSegment = insertionOrder.budget.budgetSegments.first;
```
ruben-aghayan: if we expect only one try single. it enforces that there is only one list member

lib/src/insertion_order_parser.dart
ruben-aghayan: yea +1

lib/src/insertion_order_parser.dart
```diff
@@ -166,4 +160,14 @@ class InsertionOrderParser {
         InsertionOrder_Budget_InsertionOrderAutomationType
             .INSERTION_ORDER_AUTOMATION_TYPE_UNSPECIFIED;
   }
+
+  static bool _isActiveSegment(
+      InsertionOrder_Budget_BudgetSegment segment, DateTime now) {
+    final startDate = segment.dateRange.startDate;
```
hanj7221, line 169: On the api doc: The start and end date settings of the budget segment. They are resolved relative to the parent advertiser's time zone.

Need to get the advertiser time zone from public api. 
Issue: #51

ruben-aghayan: nice catch! make sure to drop a todo with a link in the right spot in the code

lib/src/insertion_order_parser.dart
```diff
@@ -82,9 +82,13 @@ class InsertionOrderParser {
       ..budgetUnit = _createBudgetUnit(map['budgetUnit'])
       ..automationType = _createAutomationType(map['automationType']);
 
-    for (Map segmentMap in map['budgetSegments'] ?? []) {
-      budget.budgetSegments.add(_createBudgetSegment(segmentMap));
-    }
+    final segmentMaps = (map['budgetSegments'] ?? <Map<String, dynamic>>[]).cast<Map<String, dynamic>>();
```
ruben-aghayan: take a look at https://dart.dev/guides/language/effective-dart/usage#avoid-using-cast
is casting the right thing to do here?

lib/src/insertion_order_parser.dart
```diff
@@ -82,9 +82,13 @@ class InsertionOrderParser {
...
+
+    budget.activeBudgetSegment = segmentMaps
+        .map(_createBudgetSegment)
```
ruben-aghayan, line 89: nice very idiomatic!

lib/src/insertion_order_parser.dart
```diff
@@ -156,4 +160,14 @@ class InsertionOrderParser {
...
   }
+
+  static bool _isActiveSegment(
+      InsertionOrder_Budget_BudgetSegment segment, DateTime now) {
+    final startDate = segment.dateRange.startDate;
+    final endDate = segment.dateRange.endDate;
+
+    return now.isAfter(
+            DateTime(startDate.year, startDate.month, startDate.day, 0)) &&
+        now.isBefore(DateTime(endDate.year, endDate.month, endDate.day, 23, 59, 59, 999));
```
ruben-aghayan: would it be better to just say < enddate + 1 day?

lib/src/query_service.dart
```diff
@@ -2,17 +2,99 @@ import 'dart:async';
...
+    // Report start date time in milliseconds since epoch.
+    final startDateMS = DateTime(dateRange.startDate.year,
+            dateRange.startDate.month, dateRange.startDate.day, 0)
+        .millisecondsSinceEpoch;
+    final endDateMS = DateTime(dateRange.endDate.year,
+            dateRange.endDate.month, dateRange.endDate.day, 23, 59, 59, 999)
+        .millisecondsSinceEpoch;
+
+    // Report end date time in milliseconds since epoch.
+    final reportingQuery = ReportingQueryParameter()
```
ruben-aghayan: if you want, you can nest the creation of protos and make the entire query proto in one statement. i think the way you have it now is readable just raising the option

lib/src/query_service.dart
```diff
@@ -2,17 +2,99 @@ import 'dart:async';
...
+      ..params = reportingQueryParam
+      ..reportDataStartTimeMs = Int64(startDateMS)
+      ..reportDataEndTimeMs = Int64(endDateMS);
+
+    final reportingRequestArgs = RequestArgs(
+        path: reportingHttpPath,
+        method: 'POST',
+        body: reportingQuery.toProto3Json().toString());
+
+    final responseCompleter = Completer<dynamic>();
```
ruben-aghayan, line 70: may be helpful to drop a comment as to why you are using a completer

lib/src/reporting_query_parser.dart
ruben-aghayan: yea i think its technically correct but its a lot to happen in one statement. you could try putting in intermediary steps so the reader isnt tracking so many steps?

lib/src/reporting_query_parser.dart
```diff
@@ -0,0 +1,42 @@
...
+
+class ReportingQueryParser {
+  static const _emptyEntry = '';
+
+  /// Parse queryId from a json string.
+  static String parseQueryIdFromJsonString(String jsonString) =>
+      json.decode(jsonString)['queryId'] ?? _emptyEntry;
```
ruben-aghayan: is returning empty the right thing to do here? if any of these return empty then can any subsequent operations succeed?

lib/src/reporting_query_parser.dart
ruben-aghayan: nice comment

lib/src/reporting_query_parser.dart
```diff
@@ -0,0 +1,42 @@
...
+                const {})['body'] ??
+            '';
+
+    if (responseBody.isEmpty) return <String, String>{};
+
+    // Extracts the substring after 'Revenue (USD)' and before 'Report Time',
+    // and then split them on new line or comma.
+    // The resulting list should look like [ioId, revenue, ioID, revenue...].
+    final rangeStart =
+        responseBody.indexOf('Revenue (USD)') + 'Revenue (USD)'.length + 1;
```
ruben-aghayan: opt: factor revenue usd into its own variable so youre referencing the same thing when you are checking length later

lib/src/reporting_query_parser.dart
```diff
@@ -0,0 +1,42 @@
...
+    final rangeEnd = responseBody.indexOf('Report Time');
+    final stringList =
+        responseBody.substring(rangeStart, rangeEnd).split(RegExp(r'[\n,]+'));
+
+    // Make the list into a map by making ioID the key and revenue the value.
+    final revenueMap = <String, String>{};
+    for (var i = 0; i < stringList.length - 1; i = i + 2) {
```
ruben-aghayan: final i

ruben-aghayan: cool for loop

lib/src/reporting_query_parser.dart
ruben-aghayan: I think its pretty good overall

test/insertion_order_parser_test.dart
```diff
@@ -280,12 +274,12 @@ void main() {
...
           ..automationType = InsertionOrder_Budget_InsertionOrderAutomationType
               .INSERTION_ORDER_AUTOMATION_TYPE_NONE
-          ..budgetSegments.add(segment);
+          ..activeBudgetSegment = segment;
         final expected = insertionOrderTemplate..budget = budget;
         expect(actual, expected);
       });
 
-      test('multiple budgetSegments', () {
+      test('multiple budgetSegments that contains one active', () {
```
ruben-aghayan, line 285: good test

ruben-aghayan: do you have one for an unhappy path e.g. no budget segments?

ruben-aghayan (changes requested):

lib/src/insertion_order_parser.dart
hanj7221: ohh right

lib/src/excel.dart
hanj7221: Changed in commit [df81e8d](https://github.com/googleinterns/dv360-excel-plugin/pull/50/commits/df81e8dbf19a44d5365def478d7a52e807efb808#diff-5a9f11b722aad015b548cb227e5be190L132-L151)

lib/src/reporting_query_parser.dart
hanj7221: The empty checks are done [here](https://github.com/googleinterns/dv360-excel-plugin/pull/50/files#diff-fe5af7a7187ee7977d0a4f8bb86ecbafR74-R82)

test/insertion_order_parser_test.dart
hanj7221: yep its [here](https://github.com/googleinterns/dv360-excel-plugin/pull/50/files/df81e8dbf19a44d5365def478d7a52e807efb808#diff-c51d0475621a2ed583a9e1278eb92a14L235-L255)

lib/src/insertion_order_parser.dart
hanj7221: Changed to List<>.from()

lib/src/query_component.dart
```diff
@@ -31,9 +33,58 @@ class QueryComponent {
...
+  Future<Map<String, String>> _queryAndParseRevenueSpentData(
+      InsertionOrder_Budget_BudgetSegment_DateRange dateRange) async {
+    // Creates a reporting query, and parses the queryId from response.
+    final jsonCreateQueryResponse = await _queryService
+        .execReportingCreateQuery(advertiserId, insertionOrderId, dateRange);
+    final reportingQueryId = ReportingQueryParser.parseQueryIdFromJsonString(
+        stringify(jsonCreateQueryResponse));
+
+    // Uses the queryId to get the report download path.
+    if (reportingQueryId.isNotEmpty) {
```
ruben-aghayan: prefer returning early rather than nesting the rest of the method
what do you think about throwing and catching if any step in the reporting query fails?
keeping track of all the potential places empty values or null values could be from a failed request and then remembering to handle them on consumption could be confusing.

lib/src/query_service.dart
ruben-aghayan: bump?

lib/src/reporting_query_parser.dart
```diff
@@ -0,0 +1,74 @@
...
+
+    // Based on the query constructed by [QueryService._execReportingCreateQuery],
+    // the report has the following format:
+    //
+    // Header: Insertion Order ID(groupBy column), Revenue(metric column)
+    //         insertion order id, revenue
+    //         ....
+    //         row with empty string
+    //         Report parameters
+    List<List<dynamic>> reportTable = const CsvToListConverter()
```
ruben-aghayan, line 56: nice

test/reporting_query_parser_test.dart
```diff
@@ -0,0 +1,212 @@
...
+          "schedule": {
+            "frequency": "DAILY"
+          }
+        }
+        ''';
+
+        final expected =
+            ReportingQueryParser.parseDownloadPathFromJsonString(input);
+
+        final actual = 'I-am-a-download-link';
```
ruben-aghayan, line 130: lol

lib/src/reporting_query_parser.dart
hanj7221: I changed this to using the csv package now.

ruben-aghayan (approved):

lib/src/query_component.dart
```diff
@@ -31,9 +33,58 @@ class QueryComponent {
   QueryComponent(this._queryService, this._excel);
 
   void onClick() async {
-    final jsonResponse =
-        await _queryService.execQuery(advertiserId, insertionOrderId);
-    final parsedResponse = InsertionOrderParser.parse(stringify(jsonResponse));
-    await _excel.populate([parsedResponse]);
+    // Uses DV360 public APIs to fetch entity data.
```
arulprakash-g, line 36: Overall this looks good to me.

This seems focussed on single IO case. Are there batch apis we can use? What is the plan for advertiser level query?

arulprakash-g (approved):

lib/src/query_component.dart
hanj7221: For public dv3 api, we can use [this](https://developers.google.com/display-video/api/reference/rest/v1/advertisers.insertionOrders/list), which lists all the IOs under an advertiser, and we add also add a filter to let it query IOs under the same median plan. 

For reporting api, we can also batch query all the IOs under advertiser. 

Advertiser level query is coming soon!

PR #53, opened 2020-07-13
Highlight underpacing insertion orders

### LINKED ISSUE #17: Update query page UI to allow user select the underpacing option [1 day]
Besides the input boxes, display a check box where users can select if they want to highlight the underpacing insertion orders in the spreadsheet.

Cost estimate: 1 day

### LINKED ISSUE #18: Update the service class to calculate whether IO is underpacing [1~3 day]
Cost estimate: 1~3 day

### LINKED ISSUE #19: Update the service class to highlight underpacing IOs in the spreadsheet [1~2 day]
Cost estimate: 1~2 day

hanj7221:
Closes #17, closes #18, closes #19.

Here is a screenshot illustrating how a sample underpacing insertion order will be highlighted:

Screen Shot 2020-07-14 at 9 08 12 AM

### DIFF
--- a/lib/src/excel.dart
+++ b/lib/src/excel.dart
@@ -201,7 +201,7 @@ class ExcelDart {
   ///
   /// Structured reference (i.e. Table[@Column]) is not allowed in Excel
   /// conditional formatting. Regular reference (i.e. $A5) has to be used
-  /// instead. As table expands, Excel would automatically extends the
+  /// instead. As table expands, Excel will automatically extends the
   /// conditional formatting rule to cover the entire range.
   static void _highlightUnderpacingRows(String tableName) async {
     await ExcelJS.run(allowInterop((context) async {
@@ -219,6 +219,8 @@ class ExcelDart {
       // Flight duration is calculated by (end date - start date + 1), and
       // current duration is calculated by (now - start date + 1), so it is
       // guaranteed that divide by zero will not happen.
+      /// TODO: calculate difference at hours level if it is shorter than a day.
+      /// Issue: https://github.com/googleinterns/dv360-excel-plugin/issues/57
       final formula = ''' 
        ($_spentColumn/$_budgetColumn) / 
        (
--- a/lib/src/excel.dart
+++ b/lib/src/excel.dart
@@ -12,9 +12,26 @@ import 'util.dart';
 /// Service class that provides dart functions to interact with Excel.
 @Injectable()
 class ExcelDart {
-  static const _tableColumnStartAddress = 'A';
-  static const _tableRowStartAddress = 4;
+  /// Underpacing table addresses.
+  static const _underpacingTableAddress = 'A1:B2';
+  static const _underpacingHeaderAddress = 'A1:A2';
+  static const _underpacingValueAddress = 'B1:B2';
   static const _thresholdAddress = '\$B2';
+
+  /// `false` will be replaced in [ExcelDart.populate] with the correct value.
+  /// Default threshold is set to 0.9.
+  static final _underpacingTable = [
+    ['highlight underpacing:', false],
+    ['threshold:', 0.9]
+  ];
+
+  /// Excel 'bad' preset formats.
+  static const _underpacingFormatFont = '#9C0006';
+  static const _underpacingFormatFill = '#FFC7CE';
+
+  /// Result table starts at address 'A4'
+  static const _resultTableColumnStartAddress = 'A';
+  static const _resultTableRowStartAddress = 4;
 
   static const _fontName = 'Roboto';
   static const _fontSize = 12;
@@ -24,24 +41,21 @@ class ExcelDart {
   static const _leftAlignment = 'Left';
 
   static const _borderStyle = 'Continuous';
-
   static const _currencyFormat = '\$#,##0.00';
-  static const _underpacingFormatFont = '#9C0006';
-  static const _underpacingFormatFill = '#FFC7CE';
 
   /// Column addresses calculated based on [_tableHeader].
   static final _spentColumn =
       '\$${_getExcelColumnReference(_tableHeader.indexOf('Spent'))}'
-      '${_tableRowStartAddress + 1}';
+      '${_resultTableRowStartAddress + 1}';
   static final _budgetColumn =
       '\$${_getExcelColumnReference(_tableHeader.indexOf('Budget'))}'
-      '${_tableRowStartAddress + 1}';
+      '${_resultTableRowStartAddress + 1}';
   static final _startDateColumn =
       '\$${_getExcelColumnReference(_tableHeader.indexOf('Start Date'))}'
-      '${_tableRowStartAddress + 1}';
+      '${_resultTableRowStartAddress + 1}';
   static final _endDateColumn =
       '\$${_getExcelColumnReference(_tableHeader.indexOf('End Date'))}'
-      '${_tableRowStartAddress + 1}';
+      '${_resultTableRowStartAddress + 1}';
 
   /// Header for the master table.
   ///
@@ -77,26 +91,31 @@ class ExcelDart {
   void populate(List<InsertionOrder> insertionOrderList,
       bool highlightUnderpacing) async {
     await Office.onReady(allowInterop((info) async {
+      final sheetName = 'Query';
       final tableName = 'Performance_Data_Table';
 
+      /// TODO: the await here doesn't work, should use completer instead.
+      /// Issue: https://github.com/googleinterns/dv360-excel-plugin/issues/55
       await ExcelJS.run(allowInterop((context) async {
         // Adds a new worksheet.
-        final sheet = context.workbook.worksheets.add('Query');
+        final sheet = context.workbook.worksheets.add(sheetName);
 
-        // Adds and formats underpacing info.
-        sheet.getRange('Query!A1:B2')
+        // Adds and formats the underpacing table.
+        _underpacingTable[0][1] = highlightUnderpacing;
+        sheet.getRange('$sheetName!$_underpacingTableAddress')
           ..format.font.name = _fontName
           ..format.font.size = _fontSize
           ..format.font.bold = true
           ..format.horizontalAlignment = _centerAlignment
-          ..values = [
-            ['highlight underpacing:', '$highlightUnderpacing'],
-            ['threshold: ', 0.9]
-          ];
-        sheet.getRange('Query!A1:A2').format.horizontalAlignment =
-            _leftAlignment;
-        sheet.getRange('Query!B1:B2').format.horizontalAlignment =
-            _rightAlignment;
+          ..values = _underpacingTable;
+        sheet
+            .getRange('$sheetName!$_underpacingHeaderAddress')
+            .format
+            .horizontalAlignment = _leftAlignment;
+        sheet
+            .getRange('$sheetName!$_underpacingValueAddress')
+            .format
+            .horizontalAlignment = _rightAlignment;
 
         // Turns [insertionOrderList] into table rows.
         final tableBody = insertionOrderList.map(_generateTableRow).toList();
@@ -104,9 +123,10 @@ class ExcelDart {
         // Calculates table size and adds a master table.
         final endAddress =
             '${_getExcelColumnReference(_tableHeader.length - 1)}'
-            '${tableBody.length + _tableRowStartAddress}';
-        final tableAddress =
-            'Query!$_tableColumnStartAddress$_tableRowStartAddress:$endAddress';
+            '${tableBody.length + _resultTableRowStartAddress}';
+        final tableAddress = '$sheetName!'
+            '$_resultTableColumnStartAddress$_resultTableRowStartAddress:'
+            '$endAddress';
         final table = context.workbook.tables.add(tableAddress, true);
         table.name = tableName;
 
@@ -150,8 +170,8 @@ class ExcelDart {
     }));
   }
 
-  static String _getExcelColumnReference(int offset) =>
-      String.fromCharCode(_tableColumnStartAddress.codeUnitAt(0) + offset);
+  static String _getExcelColumnReference(int offset) => String.fromCharCode(
+      _resultTableColumnStartAddress.codeUnitAt(0) + offset);
 
   /// Conditionally formats the Budget column based on Budget Unit.
   ///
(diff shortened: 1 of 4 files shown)


lib/src/excel.dart
```diff
@@ -126,6 +137,32 @@ class ExcelDart {
     }));
   }
 
+  /// Highlight rows that are underpacing.
```
hanj7221, line 200: Shall I change the relative notation used in this [function](https://github.com/googleinterns/dv360-excel-plugin/pull/53/commits/847e079cc3921677bce9ae2c909ac2f2e8fbb5d8#diff-5a9f11b722aad015b548cb227e5be190R118-R138) to match the regular notation instead? So instead of referencing RC[-2], we can use $M2 to reference the budget column instead. 

Both depends on hard coding though :<

lib/src/excel.dart
```diff
@@ -19,6 +19,13 @@ class ExcelDart {
   static const _horizontalAlignment = 'Center';
   static const _borderStyle = 'Continuous';
   static const _currencyFormat = '\$#,##0.00';
+  static const _badFont = '#9C0006';
+  static const _badFill = '#FFC7CE';
+
+  static const _spentColumn = '\$N2';
```
hanj7221: Add a comment to say that this has to match with the table header & body declaration.

lib/src/excel.dart
```diff
@@ -19,6 +19,13 @@ class ExcelDart {
   static const _horizontalAlignment = 'Center';
   static const _borderStyle = 'Continuous';
   static const _currencyFormat = '\$#,##0.00';
+  static const _badFont = '#9C0006';
```
ruben-aghayan: badFill/badFont kind of vague. consider something like underpaceFont

lib/src/excel.dart
```diff
@@ -19,6 +19,13 @@ class ExcelDart {
...
+  static const _badFill = '#FFC7CE';
```
ruben-aghayan: how were these values chosen? what color is this?

lib/src/excel.dart
```diff
@@ -19,6 +19,13 @@ class ExcelDart {
...
+
+  static const _spentColumn = '\$N2';
+  static const _budgetColumn = '\$M2';
+  static const _startDateColumn = '\$O2';
+  static const _endDateColumn = '\$P2';
```
ruben-aghayan: this will all change if you add something to the schema right? is there a way to calculate this value so changes to the schema will cascade without manual updates?

lib/src/excel.dart
```diff
@@ -48,7 +55,8 @@ class ExcelDart {
   /// Waits for Office APIs to be ready and then creates
   /// a new spreadsheet with [sheetName] and populates the spreadsheet with
   /// entries in the [insertionOrderList].
-  void populate(List<InsertionOrder> insertionOrderList) async {
+  void populate(List<InsertionOrder> insertionOrderList,
+      bool highlightUnderpacing, String threshold) async {
```
ruben-aghayan: when would highlighUnderpacing be false?

ruben-aghayan: if highlight underpacing is false, does threshold do anything? do these make sense as optional arguments?

lib/src/excel.dart
ruben-aghayan: i think absolute > relative. But cant we calculate which column it is since we are specifying the schema? why hardcode?

lib/src/excel.dart
```diff
@@ -126,6 +137,32 @@ class ExcelDart {
...
+  static void _highlightUnderpacingRows(
+      String tableName, String threshold) async {
+    await ExcelJS.run(allowInterop((context) async {
+      final table = context.workbook.tables.getItem(tableName);
+      final range = table.getDataBodyRange();
+      final format = range.conditionalFormats.add('Custom');
+
+      format.custom.rule.formula = '''
+      ($_spentColumn/$_budgetColumn) *
+      (DATEDIF($_startDateColumn, $_endDateColumn, "D") /
```
ruben-aghayan: what kind of division is this? is it integer division? 
also what happens when the difference is 0?
(does the "D" mean double?) Since this is not a common library may be helpful to explain non-obvious args in a comment so they arent treated as "magic" inputs.

lib/src/query_component.dart
```diff
@@ -15,20 +15,31 @@ import 'reporting_query_parser.dart';
            placeholder="Advertiser ID: 164337" debugId="advertiser-id-input">
     <input [(ngModel)]="insertionOrderId" 
            placeholder="Insertion Order ID: 8127549" debugId="io-id-input">
+    <br>
+    <input type="checkbox" [(ngModel)]="highlightUnderpacing"
+           debugId="underpacing" name="underpacing">
+    <label for="underpacing">Highlight underpacing insertion orders</label><br>
+    <input *ngIf="highlightUnderpacing" [(ngModel)]="underpacingThreshold"
```
ruben-aghayan: this is an interesting way to do it.
an alternative would be to have it as a cell in excel. (then a user can change it and it will automatically recalculate. does that work here?)

ruben-aghayan (changes requested):
should any changes be reflected in the excel testing instructions?

lib/src/excel.dart
hanj7221: good point, good point, good point +1000.

lib/src/excel.dart
hanj7221: This mimics the 'bad' format in Excel.
Excel-bad-format

lib/src/excel.dart
hanj7221: optional arguments, yes

lib/src/excel.dart
ruben-aghayan: oh true. drop a comment to that effect!

lib/src/excel.dart
hanj7221: Actually, highlightUnderpacing will still remain as a required parameter, its value will be false if user doesn't check the checkbox. And threshold is no longer a parameter since it is changed to an excel cell input.

lib/src/excel.dart
```diff
@@ -12,13 +12,36 @@ import 'util.dart';
...
-  static const _horizontalAlignment = 'Center';
+
+  static const _centerAlignment = 'Center';
+  static const _rightAlignment = 'Right';
+  static const _leftAlignment = 'Left';
+
   static const _borderStyle = 'Continuous';
+
   static const _currencyFormat = '\$#,##0.00';
+  static const _underpacingFormatFont = '#9C0006';
```
ruben-aghayan: maybe comment that you are using the "bad" preset from excel?

lib/src/excel.dart
```diff
@@ -48,22 +71,42 @@ class ExcelDart {
...
 
+        // Adds and formats underpacing info.
+        sheet.getRange('Query!A1:B2')
+          ..format.font.name = _fontName
+          ..format.font.size = _fontSize
+          ..format.font.bold = true
+          ..format.horizontalAlignment = _centerAlignment
+          ..values = [
+            ['highlight underpacing:', '$highlightUnderpacing'],
+            ['threshold: ', 0.9]
```
ruben-aghayan: pull this out into a const

lib/src/excel.dart
```diff
@@ -12,13 +12,36 @@ import 'util.dart';
 /// Service class that provides dart functions to interact with Excel.
 @Injectable()
 class ExcelDart {
-  static const _startAddress = 'A1';
-  static final _startAddressInt = 'A'.codeUnitAt(0);
+  static const _tableColumnStartAddress = 'A';
+  static const _tableRowStartAddress = 4;
```
ruben-aghayan: how come it starts at row 4?

lib/src/excel.dart
```diff
@@ -12,13 +12,36 @@ import 'util.dart';
...
+  static const _thresholdAddress = '\$B2';
```
ruben-aghayan, line 19: can you include a screenshot? I think itll help illustrate the pr

lib/src/excel.dart
```diff
@@ -12,13 +12,36 @@ import 'util.dart';
...
+  static const _underpacingFormatFill = '#FFC7CE';
+
+  /// Column addresses calculated based on [_tableHeader].
+  static final _spentColumn =
+      '\$${_getExcelColumnReference(_tableHeader.indexOf('Spent'))}'
+      '${_tableRowStartAddress + 1}';
+  static final _budgetColumn =
+      '\$${_getExcelColumnReference(_tableHeader.indexOf('Budget'))}'
+      '${_tableRowStartAddress + 1}';
+  static final _startDateColumn =
```
ruben-aghayan, line 53: is this ok if the table doesnt exist yet (i.e. is it evaluated lazily?)

lib/src/excel.dart
```diff
@@ -126,6 +175,44 @@ class ExcelDart {
     }));
   }
 
+  /// Highlight rows that are underpacing.
+  ///
+  /// Structured reference (i.e. Table[@Column]) is not allowed in Excel
+  /// conditional formatting. Regular reference (i.e. $A5) has to be used
+  /// instead. As table expands, Excel would automatically extends the
```
ruben-aghayan: great comment
nit: will automatically extend

lib/src/excel.dart
```diff
@@ -126,6 +175,44 @@ class ExcelDart {
...
+      // (spent / budget) * (flight duration/ current duration)
+      // if current duration or flight duration is zero days, then the formula
+      // simply calculates (spent / budget).
+      final formula = '''
+      IF(OR(DATEDIF($_startDateColumn, $_endDateColumn, "D") = 0, 
+            DATEDIF($_startDateColumn, NOW(), "D") = 0),
+         ($_spentColumn/$_budgetColumn),   
+         ($_spentColumn/$_budgetColumn) * 
+         (
+           DATEDIF($_startDateColumn, $_endDateColumn, "D") /
```
ruben-aghayan: same question re:division. Is it integer division?

test/excel_test.dart
```diff
@@ -8,8 +8,14 @@
...
+/// * Can check/uncheck the highlight underpacing option.
 /// * Click on populate
 ///
 /// Expected Result:
 /// * Should see a new sheet getting created that contains a table of insertion orders.
 /// * Insertion order data values should match with those on the DV360 UI.
+/// * Should see a cell displaying the underpacing threshold value.
+/// * If highlight underpacing is checked, underpacing insertion orders should be
+///   highlighted in red.
+/// * If highlight underpacing is checked, modifying the underpacing threshold value
```
ruben-aghayan, line 23: is there a way we can test that the correct rows are being highlighted using mocks?
e.g. verify that excel.format(underPace) font was called on rows 1,2,3 but not 4,5,6?

lib/src/excel.dart
```diff
@@ -101,9 +144,15 @@ class ExcelDart {
...
+      if (highlightUnderpacing) _highlightUnderpacingRows(tableName);
     }));
   }
 
+  static String _getExcelColumnReference(int offset) =>
+      String.fromCharCode(_tableColumnStartAddress.codeUnitAt(0) + offset);
+
   /// Conditionally formats the Budget column based on Budget Unit.
   ///
   /// 'RC[-2]' in the formula follows Excel's Relative Notation and references
```
ruben-aghayan, line 178: please create issue to refactor this from hardcoded relative to calculated absolute. (also elsewhere where applicable, and drop issue here.)

ruben-aghayan (changes requested):

lib/src/excel.dart
hanj7221: Yep, sorry, that would make much more sense! See [here](https://github.com/googleinterns/dv360-excel-plugin/pull/53#issue-448467208)

lib/src/excel.dart
hanj7221: Ugh...sorry, forgot to include this in the comments. Nope, not integer division, just Excel regular division.

lib/src/excel.dart
hanj7221: sorry totally forgot to address this :(

lib/src/excel.dart
hanj7221: Please see [here](https://github.com/googleinterns/dv360-excel-plugin/pull/53#issue-448467208). Table starts on row 4, leaving some space to display underpacing options. 

If user doesn't check the underpacing checkbox, should we also display this, and shows 'highlight underpacing' as false, or should we hide this and move the table up?

lib/src/excel.dart
hanj7221: Can't turn this into a const since $highlightUnderpacing is a variable, but I will pull this out into a static final.

lib/src/excel.dart
hanj7221: Seems like so? it didn't complain or throw an error

test/excel_test.dart
hanj7221: I'm afraid not, but we can talk about this :)

lib/src/excel.dart
```diff
@@ -126,6 +197,46 @@ class ExcelDart {
...
+      // i.e. DATEDIF("7/1/2020", "7/31/2020", "D") = 30
+      //
+      // This formula calculates:
+      // (spent / budget) / (current duration / flight duration)
+      //
+      // Flight duration is calculated by (end date - start date + 1), and
+      // current duration is calculated by (now - start date + 1), so it is
+      // guaranteed that divide by zero will not happen.
+      final formula = ''' 
+       ($_spentColumn/$_budgetColumn) /
```
ruben-aghayan, line 225: not sure about this formula. lets drop an issue here and immediately follow up with another pr and get arul's eyes on it. I think adding to the numerator and denominator will change the value. Lets get better granularity than days on the date calc, and not return answers when it has been in flight for <2 days or <10% of its flight (for e.g.) to guard against weird edge cases.

lib/src/excel.dart
ruben-aghayan: bump

test/excel_test.dart
ruben-aghayan: lets drop an issue and immediately follow up. while we cant test the office api to excel ui connection, we may be able to test the excelDart to excel api connection.

ruben-aghayan (approved):

lib/src/excel.dart
ruben-aghayan: Looks good.

last thing, lets add the current pacing proportion into the schema

PR #59, opened 2020-07-14
Changing underpacing formula to consider duration less than one day

### LINKED ISSUE #57: Calculate difference at hours level if it is shorter than a day.

hanj7221:
Closes #57

### DIFF
--- a/lib/src/excel.dart
+++ b/lib/src/excel.dart
@@ -45,17 +45,8 @@ class ExcelDart {
   static const _timeFormat = '12:00:00 AM';
 
   /// Column addresses calculated based on [_tableHeader].
-  static final _spentColumn =
-      '\$${_getExcelColumnReference(_tableHeader.indexOf('Spent'))}'
-      '${_resultTableRowStartAddress + 1}';
-  static final _budgetColumn =
-      '\$${_getExcelColumnReference(_tableHeader.indexOf('Budget'))}'
-      '${_resultTableRowStartAddress + 1}';
-  static final _startDateColumn =
-      '\$${_getExcelColumnReference(_tableHeader.indexOf('Start Date'))}'
-      '${_resultTableRowStartAddress + 1}';
-  static final _endDateColumn =
-      '\$${_getExcelColumnReference(_tableHeader.indexOf('End Date'))}'
+  static final _currentPacingColumnAddress =
+      '\$${_getExcelColumnReference(_tableHeader.indexOf('Current Pacing %'))}'
       '${_resultTableRowStartAddress + 1}';
 
   /// Header for the master table.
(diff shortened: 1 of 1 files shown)


lib/src/excel.dart
```diff
@@ -216,17 +218,24 @@ class ExcelDart {
...
       // (spent / budget) / (current duration / flight duration)
       //
-      // Flight duration is calculated by (end date - start date + 1), and
-      // current duration is calculated by (now - start date + 1), so it is
-      // guaranteed that divide by zero will not happen.
-      /// TODO: calculate difference at hours level if it is shorter than a day.
-      /// Issue: https://github.com/googleinterns/dv360-excel-plugin/issues/57
+      // If current duration or flight duration is less than 1 day, then
+      // the formula calculates:
+      // (spent / budget) / (time elapsed since flight start / 24)
```
arulprakash-g: in hours

lib/src/excel.dart
```diff
@@ -216,17 +218,24 @@ class ExcelDart {
...
+      // Time elapsed has the granularity of hours and minutes.
+      // i.e. TEXT("1/1/2020 10:20:00 AM" - "1/1/2020 12:00:00 AM", "h.mm")
+      //      = 10.2
       final formula = ''' 
-       ($_spentColumn/$_budgetColumn) / 
-       (
-         (DATEDIF($_startDateColumn, NOW(), "D") + 1) /
-         (DATEDIF($_startDateColumn, $_endDateColumn, "D") + 1)
-       ) < $_thresholdAddress
+      IF(OR(DATEDIF($_startDateColumn, $_endDateColumn, "D") = 0,
```
arulprakash-g: Consider splitting this into 2. First compute pacing % & then compare against threshold. Do we expose pacing % in the sheet?

lib/src/excel.dart
```diff
@@ -216,17 +218,24 @@ class ExcelDart {
...
-       ) < $_thresholdAddress
+      IF(OR(DATEDIF($_startDateColumn, $_endDateColumn, "D") = 0, 
+            DATEDIF($_startDateColumn, NOW(), "D") = 0),
+         ($_spentColumn/$_budgetColumn) / 
+         (
+           TEXT(NOW()-$_startDateColumn, "h.mm") / 24
+         ), 
+         ($_spentColumn/$_budgetColumn) /
+         (
+           DATEDIF($_startDateColumn, NOW(), "D") /
```
arulprakash-g: Is this only for active IOs? If not (now - startDate) can be -ve for segments in future.

arulprakash-g (approved):

lib/src/excel.dart
hanj7221: Yes, we are only querying for IOs that have entity status of active. And for all the active IOs, i will also filter out those that are not in flight.

ruben-aghayan (approved):

PR #60, opened 2020-07-15
Replacing relative reference in excel formula and add completer

### LINKED ISSUE #54: Change the relative reference used in budget formatting function to absolute reference

### LINKED ISSUE #55: Use completer to turn the callback function of Excel and Office into future

hanj7221:
closes #54 and closes #55

### DIFF
--- a/lib/src/excel.dart
+++ b/lib/src/excel.dart
@@ -167,12 +167,13 @@ class ExcelDart {
       sheet.getUsedRange().getEntireColumn().format.autofitColumns();
       sheet.getUsedRange().getEntireRow().format.autofitRows();
 
-      // Sets the sheet as active and syncs context.
+      // Sets the sheet as active.
       sheet.activate();
-      context.sync();
 
       // All operations are complete.
       populateCompleter.complete();
+
+      return context.sync();
     }));
 
     // Waits for population to be complete.
@@ -206,10 +207,10 @@ class ExcelDart {
 
       format.custom.format.numberFormat = _currencyFormat;
 
-      context.sync();
-
       // All operations are complete.
       formatCompleter.complete();
+
+      return context.sync();
     }));
 
     return formatCompleter.future;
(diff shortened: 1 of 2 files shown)


lib/src/excel.dart
```diff
@@ -167,12 +167,13 @@ class ExcelDart {
...
 
-      // Sets the sheet as active and syncs context.
+      // Sets the sheet as active.
       sheet.activate();
-      context.sync();
 
       // All operations are complete.
       populateCompleter.complete();
+
+      return context.sync();
```
hanj7221, line 176: Adding back the return statement....
Excel js expects this callback function to return context.sync(). Even though the population still works without it, Excel js will complain in the console.

lib/src/query_component.dart
```diff
@@ -77,17 +80,17 @@ class QueryComponent {
...
+              json.stringify(jsonGetQueryResponse));
 
       // Downloads the report and parse the response into a revenue map.
       final report =
-      await _queryService.execReportingDownload(reportingDownloadPath);
+          await _queryService.execReportingDownload(reportingDownloadPath);
 
       return ReportingQueryParser.parseRevenueFromJsonString(report);
-    } catch(e) {
+    } catch (e) {
```
ruben-aghayan, line 93: has it been throwing? why add this now if not adding handling?

hanj7221: The try catch was added in a previous [pr](https://github.com/googleinterns/dv360-excel-plugin/pull/50#discussion_r453093375). I guess I forget to run dartfmt when I merge it, so the indentation was wrong T.T

ruben-aghayan: yea but I only just noticed it.
lets merge this for now but think about why this is here. if we are encountering this error then lets figure out why and if we arent we should let it throw so we arent swallowing errors we dont know about.

ruben-aghayan (approved):

PR #62, opened 2020-07-15
Update dv360 query to deal with advertiser-level query

### LINKED ISSUE #10: Use dv360 public api to fetch all IOs under a specific advertiser id [1~2 day]
Need to figure out how to get the list of insertion orders ids that are under a specific advertiser.

Cost estimate: 1~2 day

hanj7221:
closes #10

### DIFF
--- a/lib/src/dv360_service_response_parser.dart
+++ b/lib/src/dv360_service_response_parser.dart
@@ -2,7 +2,7 @@ import 'dart:convert';
 
 import 'proto/insertion_order_query.pb.dart';
 
-class InsertionOrderParser {
+class Dv360ServiceResponseParser {
   static const _emptyEntry = '';
 
   // Disregarding linter rule to ensure that enum maps are typed.
@@ -36,10 +36,37 @@ class InsertionOrderParser {
       key: (v) => v.name,
       value: (v) => v);
 
-  /// Parses a json string to [InsertionOrder].
-  static InsertionOrder parse(String jsonString) {
-    Map<String, dynamic> map = json.decode(jsonString);
-    return _createInsertionOrder(map);
+  /// Parses a list of [InsertionOrder] from a [jsonString] .
+  static List<InsertionOrder> parseInsertionOrders(String jsonString) {
+    Map<String, dynamic> map;
+    try {
+      map = json.decode(jsonString);
+    } catch (e) {
+      // [jsonString] is not in valid json format, returns empty list.
+      print('invalid json format');
+      return [];
+    }
+
+    return map.containsKey('insertionOrders')
+        ? List.from(map['insertionOrders'])
+            .map((ioMap) => _createInsertionOrder(ioMap))
+            .toList()
+        : [_createInsertionOrder(map)];
+  }
+
+  /// Parses the nextPageToken from a [jsonString].
+  ///
+  /// Returns empty string is input [jsonString] is null or empty.
+  static String parseNextPageToken(String jsonString) {
+    Map<String, dynamic> map;
+    try {
+      map = json.decode(jsonString);
+    } catch (e) {
+      // [jsonString] is not in valid json format, returns empty string.
+      return _emptyEntry;
+    }
+
+    return map['nextPageToken'] ?? _emptyEntry;
   }
 
   /// Creates an [InsertionOrder] instance from [map].
--- a/lib/src/public_api_parser.dart
+++ b/lib/src/public_api_parser.dart
@@ -2,7 +2,7 @@ import 'dart:convert';
 
 import 'proto/insertion_order_query.pb.dart';
 
-class Dv360ServiceResponseParser {
+class PublicApiParser {
   static const _emptyEntry = '';
 
   // Disregarding linter rule to ensure that enum maps are typed.
@@ -36,17 +36,12 @@ class Dv360ServiceResponseParser {
       key: (v) => v.name,
       value: (v) => v);
 
-  /// Parses a list of [InsertionOrder] from a [jsonString] .
+  /// Parses a list of [InsertionOrder] from a [jsonString].
+  ///
+  /// Returns an empty list if [jsonString] is null or empty.
   static List<InsertionOrder> parseInsertionOrders(String jsonString) {
-    Map<String, dynamic> map;
-    try {
-      map = json.decode(jsonString);
-    } catch (e) {
-      // [jsonString] is not in valid json format, returns empty list.
-      print('invalid json format');
-      return [];
-    }
-
+    if (jsonString == null || jsonString.isEmpty) return [];
+    Map<String, dynamic> map = json.decode(jsonString);
     return map.containsKey('insertionOrders')
         ? List.from(map['insertionOrders'])
             .map((ioMap) => _createInsertionOrder(ioMap))
@@ -56,16 +51,10 @@ class Dv360ServiceResponseParser {
 
   /// Parses the nextPageToken from a [jsonString].
   ///
-  /// Returns empty string is input [jsonString] is null or empty.
+  /// Returns a empty string if [jsonString] is null or empty.
   static String parseNextPageToken(String jsonString) {
-    Map<String, dynamic> map;
-    try {
-      map = json.decode(jsonString);
-    } catch (e) {
-      // [jsonString] is not in valid json format, returns empty string.
-      return _emptyEntry;
-    }
-
+    if (jsonString == null || jsonString.isEmpty) return _emptyEntry;
+    Map<String, dynamic> map = json.decode(jsonString);
     return map['nextPageToken'] ?? _emptyEntry;
   }
 
--- a/lib/src/public_api_parser.dart
+++ b/lib/src/public_api_parser.dart
@@ -42,7 +42,6 @@ class PublicApiParser {
   ///
   /// Returns an empty list if [jsonString] is null or empty.
   static List<InsertionOrder> parseInsertionOrders(String jsonString) {
-    if (jsonString == null || jsonString.isEmpty) return [];
     Map<String, dynamic> map = json.decode(jsonString);
 
     // If map contains key 'insertionOrders', multiple IOs are returned.
@@ -60,7 +59,6 @@ class PublicApiParser {
   ///
   /// Returns a empty string if [jsonString] is null or empty.
   static String parseNextPageToken(String jsonString) {
-    if (jsonString == null || jsonString.isEmpty) return _emptyEntry;
     Map<String, dynamic> map = json.decode(jsonString);
     return map['nextPageToken'] ?? _emptyEntry;
   }
--- a/lib/src/public_api_parser.dart
+++ b/lib/src/public_api_parser.dart
@@ -2,6 +2,8 @@ import 'dart:convert';
 
 import 'proto/insertion_order_query.pb.dart';
 
+/// TODO: parse and handle dv360 public api error.
+/// Issue:https://github.com/googleinterns/dv360-excel-plugin/issues/52
 class PublicApiParser {
   static const _emptyEntry = '';
 
@@ -42,11 +44,16 @@ class PublicApiParser {
   static List<InsertionOrder> parseInsertionOrders(String jsonString) {
     if (jsonString == null || jsonString.isEmpty) return [];
     Map<String, dynamic> map = json.decode(jsonString);
-    return map.containsKey('insertionOrders')
-        ? List.from(map['insertionOrders'])
-            .map((ioMap) => _createInsertionOrder(ioMap))
-            .toList()
-        : [_createInsertionOrder(map)];
+
+    // If map contains key 'insertionOrders', multiple IOs are returned.
+    // And if it doesn't, the map itself represents one insertion order.
+    if (map.containsKey('insertionOrders')) {
+      return List.from(map['insertionOrders'])
+          .map((ioMap) => _createInsertionOrder(ioMap))
+          .toList();
+    } else {
+      return [_createInsertionOrder(map)];
+    }
   }
 
   /// Parses the nextPageToken from a [jsonString].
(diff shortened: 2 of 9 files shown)


lib/src/query_component.dart
```diff
@@ -31,16 +39,35 @@ class QueryComponent {
...
+
+    // If [_queryType.byAdvertiser], filters out insertion orders
+    // that are not in-flight.
+    insertionOrders = _queryType == QueryType.byAdvertiser
+        ? _filterInFlightInsertionOrders(insertionOrders)
+        : insertionOrders;
+
+    // TODO: update reporting query to deal with multiple IOs.
+    // Issue: https://github.com/googleinterns/dv360-excel-plugin/issues/61
+    final insertionOrder = insertionOrders[0];
```
hanj7221: This is still querying the spend data for a single IO. Reporting api will be updated in the next pr.

lib/src/dv360_service_response_parser.dart
```diff
@@ -2,7 +2,7 @@ import 'dart:convert';
 
 import 'proto/insertion_order_query.pb.dart';
 
-class InsertionOrderParser {
+class Dv360ServiceResponseParser {
```
ruben-aghayan: nit: public api parser maybe? service response parser just sounds weird

lib/src/dv360_service_response_parser.dart
```diff
@@ -36,10 +36,37 @@ class InsertionOrderParser {
...
-    Map<String, dynamic> map = json.decode(jsonString);
-    return _createInsertionOrder(map);
+  /// Parses a list of [InsertionOrder] from a [jsonString] .
+  static List<InsertionOrder> parseInsertionOrders(String jsonString) {
+    Map<String, dynamic> map;
+    try {
+      map = json.decode(jsonString);
+    } catch (e) {
+      // [jsonString] is not in valid json format, returns empty list.
+      print('invalid json format');
```
ruben-aghayan: rm debug code

lib/src/dv360_service_response_parser.dart
```diff
@@ -36,10 +36,37 @@ class InsertionOrderParser {
...
+
+  /// Parses the nextPageToken from a [jsonString].
+  ///
+  /// Returns empty string is input [jsonString] is null or empty.
+  static String parseNextPageToken(String jsonString) {
+    Map<String, dynamic> map;
+    try {
+      map = json.decode(jsonString);
+    } catch (e) {
+      // [jsonString] is not in valid json format, returns empty string.
```
ruben-aghayan: has this been happening? do we want to swallow the error

lib/src/query_component.dart
```diff
@@ -31,16 +39,35 @@ class QueryComponent {
...
-
   bool highlightUnderpacing = false;
 
   QueryComponent(this._queryService, this._excel);
 
+  bool get isAdvertiserQuery => _queryType == QueryType.byAdvertiser;
+  set isAdvertiserQuery(bool checked) =>
+      checked ? _queryType = QueryType.byAdvertiser : null;
+
+  bool get isInsertionOrderQuery => _queryType == QueryType.byInsertionOrder;
```
ruben-aghayan: why translate your enum to a set of bools?

lib/src/query_component.dart
```diff
@@ -72,25 +116,34 @@ class QueryComponent {
...
       return <String, String>{};
     }
   }
+
+  List<InsertionOrder> _filterInFlightInsertionOrders(
+      List<InsertionOrder> insertionOrders) {
+    return insertionOrders
+        .where((io) =>
+            io.budget.activeBudgetSegment !=
+            InsertionOrder_Budget_BudgetSegment())
```
ruben-aghayan, line 152: this doesnt need to check dates?

lib/src/query_component.dart
```diff
@@ -31,16 +39,35 @@ class QueryComponent {
...
+      checked ? _queryType = QueryType.byInsertionOrder : null;
+
   void onClick() async {
     // Uses DV360 public APIs to fetch entity data.
-    final insertionOrder = await _queryAndParseInsertionOrderEntityData();
+    var insertionOrders = await _queryAndParseInsertionOrderEntityData();
+
+    // If [_queryType.byAdvertiser], filters out insertion orders
+    // that are not in-flight.
+    insertionOrders = _queryType == QueryType.byAdvertiser
```
ruben-aghayan, line 72: opt: if the filtering happened in place you could avoid re-assigning this variable

lib/src/query_component.dart
ruben-aghayan: first is more idiomatic here imo

lib/src/query_component.dart
```diff
@@ -52,16 +79,33 @@ class QueryComponent {
...
-    final response =
-        await _queryService.execDV3Query(advertiserId, insertionOrderId);
-
-    return InsertionOrderParser.parse(json.stringify(response));
+  /// then return a list of parsed [InsertionOrder] instance.
+  Future<List<InsertionOrder>> _queryAndParseInsertionOrderEntityData() async {
+    final insertionOrderList = <InsertionOrder>[];
+    var jsonResponse = '';
+
+    do {
```
ruben-aghayan, line 99: nice do while loop

lib/src/query_component.dart
```diff
@@ -52,16 +79,33 @@ class QueryComponent {
...
+      // Gets the nextPageToken, and having an empty token
+      // doesn't affect the query.
+      final nextPageToken =
+          Dv360ServiceResponseParser.parseNextPageToken(jsonResponse);
+
+      // Executes dv360 query, parses the response and adds results to the list.
+      var response = await _queryService.execDV3Query(
```
ruben-aghayan: final

lib/src/query_service.dart
```diff
@@ -114,8 +114,34 @@ class QueryService {
...
+      case QueryType.byAdvertiser:
+        {
+          final filter = 'filter=entityStatus="ENTITY_STATUS_ACTIVE"';
+          final pageToken = 'pageToken=$nextPageToken';
+          return RequestArgs(
+              path: 'https://displayvideo.googleapis.com/v1/advertisers/'
+                  '$advertiserId/insertionOrders?$filter&$pageToken',
+              method: 'GET');
+        }
+        break;
```
ruben-aghayan: do you need this break statement? is the return sufficient

test/insertion_order_parser_test.dart
```diff
@@ -41,42 +41,59 @@ void main() {
...
-      input = '{}';
-
-      final actual = InsertionOrderParser.parse(input);
-
-      final expected = InsertionOrder();
-      expect(actual, expected);
-    });
+    group('parses insertion orders from:', () {
+//      test('null', () {
+//        final actual = Dv360ServiceResponseParser.parseInsertionOrders(null);
```
ruben-aghayan: ‽

test/insertion_order_parser_test.dart
```diff
@@ -1,10 +1,10 @@
 import 'package:angular_test/angular_test.dart';
```
ruben-aghayan, line 1: update test file name to reflect class/file under test

ruben-aghayan (changes requested):

test/insertion_order_parser_test.dart
hanj7221: ugh so sorry I think I applied the wrong git stash...

lib/src/dv360_service_response_parser.dart
hanj7221: so sorry,  I think I applied the wrong git stash...

lib/src/query_component.dart
hanj7221: I haven't give much thought to how the UI will look like with different types of queries. But for developing purpose, I think a checkbox is most convenient for me now? 

It looks something like this: 
Screen Shot 2020-07-15 at 1 18 43 PM

And the bool is to have it work with the checkbox.

lib/src/dv360_service_response_parser.dart
hanj7221: In real cases, no. And in component testing with mocks, null will be passed into this function.
I guess sanity check and return empty string will be better then.

lib/src/query_component.dart
hanj7221: If the IO is not in-flight, it will have an empty InsertionOrder_Budget_BudgetSegment() for the activeBudgetSegment field.

Date checking has already been done during [parsing](https://github.com/googleinterns/dv360-excel-plugin/blob/4b50b34a1e2309e47deb360aa25076d4353a82cf/lib/src/dv360_service_response_parser.dart#L115-L118).

lib/src/query_component.dart
hanj7221: Dart doesn't seem to have an in-place filter :(

lib/src/dv360_service_response_parser.dart
ruben-aghayan: let's keep throwing. an error may be introduced and we wont notice especially because we dont have a logging framework here. later we can have some top level error handling to fail more gracefully.

lib/src/query_component.dart
ruben-aghayan: I think there should be some way to get radio buttons to work with an enum

lib/src/public_api_parser.dart
```diff
@@ -36,10 +36,26 @@ class InsertionOrderParser {
       key: (v) => v.name,
       value: (v) => v);
 
-  /// Parses a json string to [InsertionOrder].
-  static InsertionOrder parse(String jsonString) {
+  /// Parses a list of [InsertionOrder] from a [jsonString].
+  ///
+  /// Returns an empty list if [jsonString] is null or empty.
+  static List<InsertionOrder> parseInsertionOrders(String jsonString) {
+    if (jsonString == null || jsonString.isEmpty) return [];
```
ruben-aghayan: kind of concerned about the possibility of null or empty values popping up in any method. If a response was invalid i think we should fail and throw rather than propagate a null or empty string. it'll also make your life easier i think being confident in your inputs.

lib/src/public_api_parser.dart
```diff
@@ -36,10 +36,26 @@ class InsertionOrderParser {
...
     Map<String, dynamic> map = json.decode(jsonString);
-    return _createInsertionOrder(map);
+    return map.containsKey('insertionOrders')
+        ? List.from(map['insertionOrders'])
```
ruben-aghayan: meta comment this is about as big as a ternary should get from a personal readability preference.

lib/src/query_component.dart
```diff
@@ -16,9 +17,16 @@ import 'reporting_query_parser.dart';
     <input [(ngModel)]="insertionOrderId" 
            placeholder="Insertion Order ID: 8127549" debugId="io-id-input">
     <br>
+
+    <input type="checkbox" [(ngModel)]="isAdvertiserQuery" name="advertiser-query">
```
ruben-aghayan: if the query types are mutually exclusive the affordances should reflect that - typically a radio button

lib/src/query_service.dart
```diff
@@ -114,8 +114,31 @@ class QueryService {
...
 
   /// Generates query based on user inputs.
-  static String _generateQuery(String advertiserId, String insertionOrderId) {
-    return 'https://displayvideo.googleapis.com/v1/advertisers/$advertiserId/'
-        'insertionOrders/$insertionOrderId';
+  static RequestArgs _generateDV3Query(QueryType queryType,
+      String nextPageToken, String advertiserId, String insertionOrderId) {
+    switch (queryType) {
+      case QueryType.byAdvertiser:
+        {
```
ruben-aghayan: i think the inner brackets arent necessary?

ruben-aghayan (changes requested):

lib/src/public_api_parser.dart
hanj7221: I think in real cases that jsonString can never be null or empty, since even if a request fails, it still returns a json object with the error message. I'm thinking that we should add parsing to the error json and display the error to users to let them know that they have entered the wrong id, etc. I guess we should do this for all queries. 

i will create an issue and address this at end when all the queries are set. 

And the sanity check is actually just for component testing, since mock class will pass in null to this function.

lib/src/query_component.dart
```diff
@@ -45,17 +49,18 @@ class QueryComponent {
   String insertionOrderId;
   bool highlightUnderpacing = false;
 
-  QueryComponent(this._queryService, this._excel);
-
-  bool get isAdvertiserQuery => _queryType == QueryType.byAdvertiser;
-  set isAdvertiserQuery(bool checked) =>
-      checked ? _queryType = QueryType.byAdvertiser : null;
+  // Radio button states with byAdvertiser selected as default.
+  RadioButtonState advertiserQuery = RadioButtonState(true, 'byAdvertiser');
```
hanj7221, line 53: Seems like the radio button can only take string values :(

Saw [other people](https://github.com/dart-lang/angular/issues/293) complain about this too.

lib/src/public_api_parser.dart
ruben-aghayan: You can set the mock class to return a default value in setUp. i think it's better to facilitate testing needs in the test code rather than your business code

lib/src/query_service.dart
ruben-aghayan: looks like these are still here

ruben-aghayan (changes requested):

lib/src/public_api_parser.dart
ruben-aghayan: wanted to reply to this as part of a review to draw your attention to it

ruben-aghayan (changes requested):

lib/src/query_service.dart
hanj7221: Ohhh..I thought i can only get rid of brackets if the statement is one line.

lib/src/public_api_parser.dart
hanj7221: Ohh right. I will just get rid of the sanity checks then. The TODO for adding parsing support to error is at the top of the file.

ruben-aghayan (approved):

PR #64, opened 2020-07-16
Update reporting api for advertiser-level query

### LINKED ISSUE #61: Update reporting query to deal with multiple insertion orders

hanj7221:
Closes #61

### DIFF
--- a/lib/src/query_component.dart
+++ b/lib/src/query_component.dart
@@ -38,7 +38,7 @@ import 'util.dart';
   ],
   directives: [coreDirectives, formDirectives],
 )
-class QueryComponent {
+class QueryComponent implements OnInit {
   final buttonName = 'populate';
   final QueryService _queryService;
   final ExcelDart _excel;
@@ -54,6 +54,9 @@ class QueryComponent {
   RadioButtonState insertionOrderQuery = RadioButtonState(false, 'byIO');
 
   QueryComponent(this._queryService, this._excel);
+
+  @override
+  void ngOnInit() async => await _excel.loadOffice();
 
   void onClick() async {
     // Determines the query type from radio buttons.
--- a/lib/src/query_component.dart
+++ b/lib/src/query_component.dart
@@ -1,10 +1,12 @@
 import 'package:angular/angular.dart';
 import 'package:angular_forms/angular_forms.dart';
+import 'package:quiver/collection.dart';
 
 import 'excel.dart';
-import 'public_api_parser.dart';
+import 'insertion_order_daily_spend.dart';
 import 'json_js.dart';
 import 'proto/insertion_order_query.pb.dart';
+import 'public_api_parser.dart';
 import 'query_service.dart';
 import 'reporting_api_parser.dart';
 import 'util.dart';
@@ -115,16 +117,16 @@ class QueryComponent implements OnInit {
 
   /// Fetches revenue spent data using DBM reporting APIs,
   /// then return a map with <ioID, revenue> key-value pairs.
-  Future<Map<String, List<InsertionOrderDailySpend>>> _queryAndParseSpentData(
+  Future<Multimap<String, InsertionOrderDailySpend>> _queryAndParseSpentData(
       DateTime startDate, DateTime endDate) async {
-    // Creates a reporting query, and parses the queryId from response.
-    final jsonCreateQueryResponse =
-        await _queryService.execReportingCreateQuery(
-            _queryType, advertiserId, insertionOrderId, startDate, endDate);
-    final reportingQueryId = ReportingQueryParser.parseQueryIdFromJsonString(
-        JsonJS.stringify(jsonCreateQueryResponse));
-
     try {
+      // Creates a reporting query, and parses the queryId from response.
+      final jsonCreateQueryResponse =
+          await _queryService.execReportingCreateQuery(
+              _queryType, advertiserId, insertionOrderId, startDate, endDate);
+      final reportingQueryId = ReportingQueryParser.parseQueryIdFromJsonString(
+          JsonJS.stringify(jsonCreateQueryResponse));
+
       // Uses the queryId to get the report download path.
       final jsonGetQueryResponse =
           await _queryService.execReportingGetQuery(reportingQueryId);
@@ -140,7 +142,7 @@ class QueryComponent implements OnInit {
     } catch (e) {
       /// TODO: proper error handling.
       /// Issue: https://github.com/googleinterns/dv360-excel-plugin/issues/52.
-      return <String, List<InsertionOrderDailySpend>>{};
+      return Multimap<String, InsertionOrderDailySpend>();
     }
   }
 
@@ -161,11 +163,8 @@ class QueryComponent implements OnInit {
           .reduce((date, minDate) => date.isBefore(minDate) ? date : minDate);
 
   void _addSpentToInsertionOrders(List<InsertionOrder> insertionOrders,
-      Map<String, List<InsertionOrderDailySpend>> spendingMap) {
+      Multimap<String, InsertionOrderDailySpend> spendingMap) {
     for (final io in insertionOrders) {
-      // Continue if the spending map doesn't have spend data for this io.
-      if (!spendingMap.containsKey(io.insertionOrderId)) continue;
-
       final budgetUnit = io.budget.budgetUnit;
       final flightStart = Util.convertProtoDateToDateTime(
           io.budget.activeBudgetSegment.dateRange.startDate);
--- a/lib/src/query_component.dart
+++ b/lib/src/query_component.dart
@@ -6,7 +6,7 @@ import 'public_api_parser.dart';
 import 'json_js.dart';
 import 'proto/insertion_order_query.pb.dart';
 import 'query_service.dart';
-import 'reporting_query_parser.dart';
+import 'reporting_api_parser.dart';
 import 'util.dart';
 
 @Component(
@@ -70,18 +70,17 @@ class QueryComponent {
         ? _filterInFlightInsertionOrders(insertionOrders)
         : insertionOrders;
 
-    // TODO: update reporting query to deal with multiple IOs.
-    // Issue: https://github.com/googleinterns/dv360-excel-plugin/issues/61
-    final insertionOrder = insertionOrders.first;
+    // Gets the earliest start date from the list of insertion orders.
+    final minStartDate = _getMinStartDate(insertionOrders);
 
-    // Gets dateRange for the active budget segment.
-    final activeDateRange = insertionOrder.budget.activeBudgetSegment.dateRange;
+    // Uses DBM reporting APIs to get spent (currency based or
+    // impression based) within time window [minStartDate, Now].
+    // spendingMap has a [DailySpend] for each insertion order.
+    final spendingMap =
+        await _queryAndParseSpentData(minStartDate, DateTime.now());
 
-    // Uses DBM reporting APIs to get revenue data within [activeDateRange].
-    final revenueMap = await _queryAndParseRevenueSpentData(activeDateRange);
-
-    // Add revenue spent to the parsed Insertion Order.
-    insertionOrder.spent = revenueMap[insertionOrder.insertionOrderId] ?? '';
+    // Add spent to the list of insertion orders.
+    _addSpentToInsertionOrders(insertionOrders, spendingMap);
 
     // Populate the spreadsheet.
     await _excel.populate(insertionOrders, highlightUnderpacing);
@@ -113,11 +112,12 @@ class QueryComponent {
 
   /// Fetches revenue spent data using DBM reporting APIs,
   /// then return a map with <ioID, revenue> key-value pairs.
-  Future<Map<String, String>> _queryAndParseRevenueSpentData(
-      InsertionOrder_Budget_BudgetSegment_DateRange dateRange) async {
+  Future<Map<String, List<InsertionOrderDailySpend>>> _queryAndParseSpentData(
+      DateTime startDate, DateTime endDate) async {
     // Creates a reporting query, and parses the queryId from response.
-    final jsonCreateQueryResponse = await _queryService
-        .execReportingCreateQuery(advertiserId, insertionOrderId, dateRange);
+    final jsonCreateQueryResponse =
+        await _queryService.execReportingCreateQuery(
+            _queryType, advertiserId, insertionOrderId, startDate, endDate);
     final reportingQueryId = ReportingQueryParser.parseQueryIdFromJsonString(
         JsonJS.stringify(jsonCreateQueryResponse));
 
@@ -137,7 +137,7 @@ class QueryComponent {
     } catch (e) {
       /// TODO: proper error handling.
       /// Issue: https://github.com/googleinterns/dv360-excel-plugin/issues/52.
-      return <String, String>{};
+      return <String, List<InsertionOrderDailySpend>>{};
     }
   }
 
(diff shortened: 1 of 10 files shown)


lib/src/util.dart
```diff
@@ -15,4 +18,24 @@ class Util {
...
+    return DateFormat('yyyy/MM/dd').parse(date);
+  }
+
+  /// Return `true` if [target] is before [start] and after [end],
+  /// false otherwise.
+  static bool isBetweenDates(DateTime target, DateTime start, DateTime end) =>
+      (target.isAfter(start) || target.isAtSameMomentAs(start)) &&
+      (target.isBefore(end) || target.isAtSameMomentAs(end));
+
+  /// Add two string values together.
```
hanj7221, line 41: Not sure if there is precision loss here. The returned spent from reporting api is a string and the value is not in micros.

test/public_api_parser_test.dart
```diff
@@ -42,20 +42,6 @@ void main() {
     tearDown(disposeAnyRunningTest);
 
     group('parses insertion orders from:', () {
-      test('null', () {
```
hanj7221, line 45: No longer doing sanity checks. Shouldn't encounter these inputs in real cases.

lib/src/reporting_api_parser.dart
```diff
@@ -30,8 +32,12 @@ class ReportingQueryParser {
...
 
-  /// Parse revenue from a json string.
-  static Map<String, String> parseRevenueFromJsonString(String jsonString) {
+  /// Parses spending information from a json string.
+  ///
+  /// Returns a spendingMap that has insertion order as key and
+  /// [InsertionOrderDailySpend] as value.
+  static Map<String, List<InsertionOrderDailySpend>> parseRevenueFromJsonString(
+      String jsonString) {
     if (jsonString == null || jsonString.isEmpty || jsonString == 'null') {
```
hanj7221: Oops, should get rid of this sanity check.

test/reporting_api_parser_test.dart
```diff
@@ -183,29 +184,68 @@ void main() {
...
-          '333333': '0.00'
-        };
-        expect(actual, expected);
+        final firstSpending = actual['111111'];
+        final secondSpending = actual['222222'];
+        final thirdSpending = actual['333333'];
+
+        expect(actual is Map<String, List<InsertionOrderDailySpend>>, true);
+
+        expect(firstSpending is List<InsertionOrderDailySpend>, true);
```
ruben-aghayan: you could create your own expected proto and expect equivalence. this would compare all the fields at the same time.

lib/src/query_component.dart
```diff
@@ -73,18 +73,17 @@ class QueryComponent implements OnInit {
...
+    final spendingMap =
+        await _queryAndParseSpentData(minStartDate, DateTime.now());
 
-    // Uses DBM reporting APIs to get revenue data within [activeDateRange].
-    final revenueMap = await _queryAndParseRevenueSpentData(activeDateRange);
-
-    // Add revenue spent to the parsed Insertion Order.
-    insertionOrder.spent = revenueMap[insertionOrder.insertionOrderId] ?? '';
+    // Add spent to the list of insertion orders.
+    _addSpentToInsertionOrders(insertionOrders, spendingMap);
```
ruben-aghayan, line 88: i like this method name because it makes it clear the operation is in place. nice

lib/src/query_component.dart
```diff
@@ -116,11 +115,12 @@ class QueryComponent implements OnInit {
 
   /// Fetches revenue spent data using DBM reporting APIs,
   /// then return a map with <ioID, revenue> key-value pairs.
-  Future<Map<String, String>> _queryAndParseRevenueSpentData(
-      InsertionOrder_Budget_BudgetSegment_DateRange dateRange) async {
+  Future<Map<String, List<InsertionOrderDailySpend>>> _queryAndParseSpentData(
```
ruben-aghayan: is there an existing multimap data type you could use? quiver lib has some more advanced data types like multimaps

lib/src/query_component.dart
```diff
@@ -152,4 +152,36 @@ class QueryComponent implements OnInit {
...
+      insertionOrders
+          .map((io) => Util.convertProtoDateToDateTime(
+              io.budget.activeBudgetSegment.dateRange.startDate))
+          .reduce((date, minDate) => date.isBefore(minDate) ? date : minDate);
+
+  void _addSpentToInsertionOrders(List<InsertionOrder> insertionOrders,
+      Map<String, List<InsertionOrderDailySpend>> spendingMap) {
+    for (final io in insertionOrders) {
+      // Continue if the spending map doesn't have spend data for this io.
+      if (!spendingMap.containsKey(io.insertionOrderId)) continue;
```
ruben-aghayan: under what circumstances do we expect this data to be missing?

lib/src/util.dart
```diff
@@ -15,4 +18,24 @@ class Util {
...
+
+  /// Convert a string date(yyyy/mm/dd) to Dart DateTime.
+  static DateTime convertStringDateToDateTime(String date) {
+    return DateFormat('yyyy/MM/dd').parse(date);
+  }
+
+  /// Return `true` if [target] is before [start] and after [end],
+  /// false otherwise.
+  static bool isBetweenDates(DateTime target, DateTime start, DateTime end) =>
+      (target.isAfter(start) || target.isAtSameMomentAs(start)) &&
```
ruben-aghayan, line 38: good consideration of edge cases.
this may be a good place to drop a bug indication that we are not handling timezones. Are the start/stop timezones at the same timezone as reporting api?

lib/src/util.dart
ruben-aghayan: This is a good callout. consider adding it to the doc comment.

lib/src/util.dart
```diff
@@ -15,4 +18,24 @@ class Util {
...
+      (target.isBefore(end) || target.isAtSameMomentAs(end));
+
+  /// Add two string values together.
+  static String addStringRevenue(String revenueA, String revenueB) =>
+      (double.parse(revenueA) + double.parse(revenueB)).toString();
```
ruben-aghayan: can you use an int64 here instead? typically we avoid using floats to represent money

lib/src/util.dart
```diff
@@ -1,4 +1,7 @@
 import 'package:fixnum/fixnum.dart';
+import 'package:intl/intl.dart';
```
ruben-aghayan, line 3: where is intl used?

lib/src/reporting_api_parser.dart
```diff
@@ -59,16 +66,36 @@ class ReportingQueryParser {
...
+          insertionOrderId, () => <InsertionOrderDailySpend>[]);
+      dailySpendList.add(InsertionOrderDailySpend(date, revenue, impression));
+      revenueMap[insertionOrderId] = dailySpendList;
     }
 
     return revenueMap;
   }
 }
+
+class InsertionOrderDailySpend {
```
ruben-aghayan: this looks like a good opportunity to use a builtclass
https://pub.dev/documentation/built_value/latest/built_value/Built-class.html

its helps build data classes in dart.

lib/src/query_service.dart
```diff
@@ -135,4 +108,47 @@ class QueryService {
...
+        ..filters.add(ReportingQueryParameter_Params_Filters()
+          ..type = '"FILTER_ADVERTISER"'
+          ..value = advertiserId))
+      ..reportDataStartTimeMs = Int64(startDate.millisecondsSinceEpoch)
+      ..reportDataEndTimeMs = Int64(endDate.millisecondsSinceEpoch);
+
+    switch (queryType) {
+      case QueryType.byAdvertiser:
+        break;
+
```
ruben-aghayan, line 138: is there an extra space here?

lib/src/query_service.dart
```diff
@@ -135,4 +108,47 @@ class QueryService {
         return RequestArgs();
     }
   }
+
+  /// Generates query based on user inputs.
+  static RequestArgs _generateReportingQuery(
```
ruben-aghayan, line 113: this is a good refactor

lib/src/reporting_api_parser.dart
```diff
@@ -59,16 +66,36 @@ class ReportingQueryParser {
...
       // stop if reach the empty string row
       if (row[0].isEmpty) return revenueMap;
 
-      revenueMap[row[0]] = row[1];
+      final insertionOrderId = row[0];
+      final date = Util.convertStringDateToDateTime(row[1]);
+      final revenue = row[2];
+      final impression = row[3];
+
+      final dailySpendList = revenueMap.putIfAbsent(
```
ruben-aghayan: i still love pythons default dictionary

lib/src/query_component.dart
```diff
@@ -152,4 +152,36 @@ class QueryComponent implements OnInit {
...
+          io.budget.activeBudgetSegment.dateRange.endDate);
+
+      final spent = spendingMap[io.insertionOrderId]
+          .where((dailySpend) =>
+              Util.isBetweenDates(dailySpend.date, flightStart, flightEnd))
+          .map((dailySpend) => budgetUnit ==
+                  InsertionOrder_Budget_BudgetUnit.BUDGET_UNIT_CURRENCY
+              ? dailySpend.revenue
+              : dailySpend.impression)
+          .reduce((spend, sum) => Util.addStringRevenue(spend, sum));
```
ruben-aghayan, line 181: this way you will end up iterating over the data twice (once to build the map and once to see which values should be added). Imo i think this is the right way to do it because the division of responsibility and the code readability is more important.

ruben-aghayan (changes requested):

lib/src/query_component.dart
hanj7221: If the IO is active and inflight, but the reporting api doesn't have spending data for it? Sounds like this is not really possible, should we get rid of the sanity check then?

lib/src/util.dart
hanj7221: I think it is not an integer value. Does [decimal](https://pub.dev/packages/decimal) sound like a better choice?

lib/src/util.dart
hanj7221: Reporting api takes ms since unix epoch as input, so I think the target/start/end here are all at the same timezones since I converted the flight start/end to ms and then pass them to the reporting api. 

but isBetweenDates(DateTime.now(), flight start, flight end) will have issues, since DateTime.now() returns current date and time in the local time zone, but public api returns flight start/end in advertiser time zone.

I'll add a TODO and link the issue.

lib/src/util.dart
hanj7221: yea the names are totally not related. DateFormat uses it to [parse string dates into DateTime](https://github.com/googleinterns/dv360-excel-plugin/blob/7e4b88f8be83899337522a9b8b62f178bf710882/lib/src/util.dart#L29).

lib/src/query_component.dart
hanj7221: wooo me likey

lib/src/query_service.dart
hanj7221: Line 138? Is the line between two cases optional?

lib/src/util.dart
hanj7221: How about [decimal](https://pub.dev/packages/decimal), do you think there still might be precision loss there?

ruben-aghayan: Decimal should be good!

lib/src/query_component.dart
ruben-aghayan: if we dont expect this case to happen, I think we should validate the response when we receive it. You can leave this here until we add validation on the service? up to you.

lib/src/query_service.dart
ruben-aghayan: my mistake. I saw two new lines for some reason.

lib/src/query_component.dart
```diff
@@ -152,4 +154,33 @@ class QueryComponent implements OnInit {
             InsertionOrder_Budget_BudgetSegment())
         .toList();
   }
+
+  DateTime _getMinStartDate(
+          List<InsertionOrder> insertionOrders) =>
+      insertionOrders
+          .map((io) => Util.convertProtoDateToDateTime(
```
ruben-aghayan, line 161: opt: consider doing this backwards (filter then convert). This is a minor optimization so readability is more important if you think this is better.

lib/src/reporting_api_parser.dart
```diff
@@ -0,0 +1,64 @@
...
+
+  /// Parses spending information from a json string.
+  ///
+  /// Returns a spendingMap that has insertion order as key and
+  /// [InsertionOrderDailySpend] as value.
+  static Multimap<String, InsertionOrderDailySpend> parseRevenueFromJsonString(
+      String jsonString) {
+    String report = json.decode(jsonString)['gapiRequest']['data']['body'];
+
+    // Based on the query constructed by [QueryService._execReportingCreateQuery],
```
ruben-aghayan, line 27: nice comment

lib/src/reporting_api_parser.dart
```diff
@@ -0,0 +1,64 @@
...
+    // and then put insertion order and the other entries into a map
+    // by making ioID the key and InsertionOrderDailySpend the value.
+    final revenueMap = Multimap<String, InsertionOrderDailySpend>();
+    for (final row in reportTable) {
+      // skip header
+      if (row[0] == 'Insertion Order ID') continue;
+      // stop if reach the empty string row
+      if (row[0].isEmpty) return revenueMap;
+
+      final insertionOrderId = row[0];
```
ruben-aghayan: opt: could probably do this inline when building the proto

test/reporting_api_parser_test.dart
```diff
@@ -0,0 +1,206 @@
...
+
+      test('a report json string that contains a single row of revenue value',
+          () {
+        final reportBody = 'Insertion Order ID, Date, Revenue, Impression\\n'
+            '123456,2020/01/01,88.88,1000\\n';
+        input = generateInput(reportBody);
+
+        final actual = ReportingQueryParser.parseRevenueFromJsonString(input);
+
+        final expectedSpending = InsertionOrderDailySpend((b) => b
```
ruben-aghayan: is a builder generated for you?

test/reporting_api_parser_test.dart
```diff
@@ -0,0 +1,206 @@
...
+          ..date = DateTime(2020, 1, 1)
+          ..revenue = '88.88'
+          ..impression = '1000');
+
+        expect(actual is Multimap<String, InsertionOrderDailySpend>, true);
```
ruben-aghayan: dont need to check type if its in the signature of the method

test/reporting_api_parser_test.dart
```diff
@@ -0,0 +1,206 @@
...
+        expect(actual.length, 1);
+        expect(actual['123456'].length, 1);
```
ruben-aghayan: 143 will fail if this is not true

test/reporting_api_parser_test.dart
```diff
@@ -0,0 +1,206 @@
...
+
+        final actual = ReportingQueryParser.parseRevenueFromJsonString(input);
+
+        final expectedSpending = InsertionOrderDailySpend((b) => b
+          ..date = DateTime(2020, 1, 1)
+          ..revenue = '88.88'
+          ..impression = '1000');
+
+        expect(actual is Multimap<String, InsertionOrderDailySpend>, true);
+        expect(actual.length, 1);
```
ruben-aghayan: use hasLength() matcher

ruben-aghayan (changes requested):

test/reporting_api_parser_test.dart
hanj7221: That variable "b" stands for builder. Is there a better syntax for builder? I'm following the example [here](https://github.com/google/built_value.dart/blob/master/example/lib/example.dart).

I guess I can also do it like this:

`final spending = InsertionOrderDailySpendBuilder..date = date..revenue = revenue..impression = impression.build() 
`


Do you think that will be better?

test/reporting_api_parser_test.dart
hanj7221: Oooo, 142 and 143 seems redundant, seems like checking the length here is not necessary?

hanj7221: I copied the mapEqual function over and make it work for multiMap as well. No longer doing these checks now.

test/reporting_api_parser_test.dart
hanj7221: I copied the mapEqual function over and make it work for multiMap as well. No longer doing these checks now.

lib/src/query_component.dart
hanj7221: The second part is a reduce instead of filter, so I thinking keeping it as it was might be better?

test/reporting_api_parser_test.dart
ruben-aghayan: if the lib documentation suggests that format then its fair game. my personal preference is using the builder; it's more readable imo. Up to you.

test/reporting_api_parser_test.dart
```diff
@@ -0,0 +1,220 @@
...
+        }
+        ''';
+      }
+
+      setUp(() => expected = Multimap<String, InsertionOrderDailySpend>());
+
+      // Copies the MapEqual function.
+      // Replaces`bValue == a[k]` in the original function with
+      // `listsEqual(bValue, a[k]`
+      bool multiMapEqual(Multimap a, Multimap b) {
```
ruben-aghayan: opt: make a custom matcher and use that instead

ruben-aghayan (approved):

PR #65, opened 2020-07-21
Implementing query by media plan and testing query component

### LINKED ISSUE #13: Update the service class to also store the media plan id [1 day]
Cost estimate: 1 day

### LINKED ISSUE #14: Use dv360 api to fetch all IOs under (advertiser id + media plan id) [1 day]
Cost estimate: 1 day

hanj7221:
closes #13 and closes #14 

Here is how the UI now looks like:

Screen Shot 2020-07-21 at 4 34 13 PM


lib/src/query_component.dart
```diff
@@ -1,10 +1,11 @@
+import 'dart:html';
```
hanj7221: Todo: delete this import

lib/src/query_component.dart
```diff
@@ -45,35 +49,31 @@ class QueryComponent implements OnInit {
   final QueryService _queryService;
   final ExcelDart _excel;
 
-  QueryType _queryType;
+  List<QueryType> queryTypeChoices = QueryType.values;
+  QueryType queryType;
 
-  String advertiserId;
-  String insertionOrderId;
+  String advertiserId = '164337';
```
ruben-aghayan: why these default values? i think it is best to leave it blank, it may not be ok to expose these ids publicly.

lib/src/query_component.dart
```diff
@@ -16,17 +17,20 @@ import 'util.dart';
   template: '''
```
ruben-aghayan: the template is complex enough to have earned its own file

lib/src/query_component.dart
```diff
@@ -45,35 +49,31 @@ class QueryComponent implements OnInit {
...
   void ngOnInit() async => await _excel.loadOffice();
 
   void onClick() async {
-    // Determines the query type from radio buttons.
-    _queryType = advertiserQuery.checked
-        ? QueryType.byAdvertiser
-        : QueryType.byInsertionOrder;
-
     // Uses DV360 public APIs to fetch entity data.
     var insertionOrders = await _queryAndParseInsertionOrderEntityData();
```
ruben-aghayan: opt: add lhs data type. it's up to taste. Imo it's not obvious just by reading the method name (unlike a contructor for example) even though its inferable from the signature.

lib/src/query_component.dart
```diff
@@ -45,35 +49,31 @@ class QueryComponent implements OnInit {
...
 
-    // If [_queryType.byAdvertiser], filters out insertion orders
+    // Early return if the query returns no insertion orders.
+    if (insertionOrders.isEmpty) return;
+
+    // If query type is not by insertion order, filters out insertion orders
     // that are not in-flight.
-    insertionOrders = _queryType == QueryType.byAdvertiser
```
ruben-aghayan, line 74: no longer filtering queries byAdvertiser?

lib/src/query_service.dart
```diff
@@ -16,14 +17,14 @@ class QueryService {
...
+      String advertiserId, String mediaPlanId, String insertionOrderId) async {
+    final responseCompleter = Completer<String>();
     GoogleAPI.client
-        .request(_generateDV3Query(
-            queryType, nextPageToken, advertiserId, insertionOrderId))
+        .request(_generateDV3Query(queryType, nextPageToken, advertiserId,
+            mediaPlanId, insertionOrderId))
         .execute(allowInterop((jsonResp, rawResp) {
-      responseCompleter.complete(jsonResp);
+      responseCompleter.complete(JsonJS.stringify(jsonResp));
```
ruben-aghayan, line 27: nice.

lib/src/query_service.dart
```diff
@@ -87,15 +89,29 @@ class QueryService {
...
-  static RequestArgs _generateDV3Query(QueryType queryType,
-      String nextPageToken, String advertiserId, String insertionOrderId) {
+  static RequestArgs _generateDV3Query(
+      QueryType queryType,
+      String nextPageToken,
+      String advertiserId,
+      String mediaPlanId,
+      String insertionOrderId) {
+    const entityStatusFilter = 'filter=entityStatus="ENTITY_STATUS_ACTIVE"';
+    final mediaPlanFilter = 'filter=campaignId="$mediaPlanId"';
```
ruben-aghayan: is this used outside the switch for query by media plan?

lib/src/query_service.dart
```diff
@@ -136,6 +153,12 @@ class QueryService {
       case QueryType.byAdvertiser:
         break;
 
+      case QueryType.byMediaPlan:
+        parameter.params.filters.add(ReportingQueryParameter_Params_Filters()
```
ruben-aghayan, line 157: nice

lib/src/query_component.dart
```diff
@@ -16,17 +17,20 @@ import 'util.dart';
   template: '''
     <input [(ngModel)]="advertiserId" 
            placeholder="Advertiser ID: 164337" debugId="advertiser-id-input">
+    <input [(ngModel)]="mediaPlanId"
```
ruben-aghayan: can you add a screenshot so to visualize the template changes?

ruben-aghayan (changes requested):

lib/src/query_component.dart
hanj7221: My mistake, forgot to delete them when pushing :( 
I guess I will do squash & merge this time.

lib/src/query_component.dart
hanj7221: I flipped the logic so it it not filtering when querying byInsertionOrder, but filtering when byAdvertiser/byMediaPlan

lib/src/query_service.dart
hanj7221: I guessed not, moving it to the inside then.

lib/src/query_component.dart
hanj7221: Of course, please see [here](https://github.com/googleinterns/dv360-excel-plugin/pull/65#issue-454717402).

lib/src/query_component.dart
ruben-aghayan: oh nice.

lib/src/query_component.html
```diff
@@ -0,0 +1,17 @@
+<input [(ngModel)]="advertiserId"
+       placeholder="Advertiser ID" debugId="advertiser-id-input">
+<input [(ngModel)]="mediaPlanId"
+       placeholder="Media Plan ID" debugId="media-plan-id-input">
+<input [(ngModel)]="insertionOrderId"
+       placeholder="Insertion Order ID" debugId="io-id-input"><br>
+<div *ngFor="let choice of queryTypeChoices">
+    <input type="radio" (click)="queryType = choice" name="radio-group"
+           id="{{enum2String(choice)}}-radio-btn"/>
```
ruben-aghayan: right now the user will be seeing camelCase var name as the button label. Is there a way to display a user-friendly string here?

lib/src/query_component.html
```diff
@@ -0,0 +1,17 @@
...
+    <label>{{enum2String(choice)}}</label><br>
+</div>
+<input type="checkbox" [(ngModel)]="highlightUnderpacing"
+       debugId="underpacing-checkbox" name="underpacing">
+<label>Highlight underpacing insertion orders</label><br>
+<button (click)="onClick()" debugId="populate-btn">
+    {{buttonName}}
+</button>
```
ruben-aghayan: newline

lib/src/query_component.html
```diff
@@ -0,0 +1,17 @@
...
+<input [(ngModel)]="insertionOrderId"
+       placeholder="Insertion Order ID" debugId="io-id-input"><br>
+<div *ngFor="let choice of queryTypeChoices">
+    <input type="radio" (click)="queryType = choice" name="radio-group"
+           id="{{enum2String(choice)}}-radio-btn"/>
+    <label>{{enum2String(choice)}}</label><br>
+</div>
+<input type="checkbox" [(ngModel)]="highlightUnderpacing"
+       debugId="underpacing-checkbox" name="underpacing">
+<label>Highlight underpacing insertion orders</label><br>
```
ruben-aghayan: this should be defined in dart

lib/src/query_component.html
```diff
@@ -0,0 +1,17 @@
...
+<button (click)="onClick()" debugId="populate-btn">
+    {{buttonName}}
```
ruben-aghayan: opt: more descriptive var name for button name. like populateButtonName

test/query_component_test.dart
```diff
@@ -72,17 +77,137 @@ void main() {
...
+        'selecting byAdvertiser invokes execDV3Query(Query.byAdvertiser, '
+        ', null, null, null)', () async {
+      await (queryComponentPO.selectByAdvertiser());
+      await (queryComponentPO.clickPopulate());
+      verify(mockQueryService.execDV3Query(
+          QueryType.byAdvertiser, '', null, null, null));
+    });
+
+    test(
+        'selecting byAdvertiser invokes execDV3Query(Query.byAdvertiser, '
```
ruben-aghayan: nit: name

test/query_component_test.dart
```diff
@@ -72,17 +77,137 @@ void main() {
...
+        ', null, null, null)', () async {
+      await (queryComponentPO.selectByIO());
+      await (queryComponentPO.clickPopulate());
+      verify(mockQueryService.execDV3Query(
+          QueryType.byInsertionOrder, '', null, null, null));
+    });
+
+    test('only the last selection matters', () async {
```
ruben-aghayan: great test!

test/query_component_test.dart
```diff
@@ -72,17 +77,137 @@ void main() {
...
+      await (queryComponentPO.selectByIO());
+      await (queryComponentPO.selectByAdvertiser());
+      await (queryComponentPO.selectByMediaPlan());
+      await (queryComponentPO.clickPopulate());
+      verify(mockQueryService.execDV3Query(
```
ruben-aghayan: what if i hit populate without making a radio button selection? is there a default value?

test/query_component_test.dart
```diff
@@ -72,17 +77,137 @@ void main() {
...
+          QueryType.byMediaPlan, '', null, null, null));
+    });
   });
 
-  test('populate button clicks invokes ExcelDart.populate()', () async {
-    await queryComponentPO.clickPopulate();
-    verify(mockExcelDart.populate(any, any));
+  group('populate-button clicks invokes functions:', () {
```
ruben-aghayan: maybe a better test would be to execQuery and expect on the output. This way if you change your implementation, only the setup will have to be changed not the test itself. Test behavior not implementation. Is there an output on execQuery we can observe instead?

Note: this is different from testing interaction w/ office APIs b/c we control the implementation here.

test/testing/query_component_po.dart
```diff
@@ -11,5 +11,43 @@ abstract class QueryComponentPageObject {
...
+  @ByDebugId('advertiser-id-input')
+  PageLoaderElement get _advertiserIdInputBox;
+
+  @ByDebugId('media-plan-id-input')
+  PageLoaderElement get _mediaPlanIdInputBox;
+
+  @ByDebugId('io-id-input')
+  PageLoaderElement get _insertionOrderIdInputBox;
+
+  @ById('byAdvertiser-radio-btn')
```
ruben-aghayan: opt: a radio button po that has the buttons inside it.

ruben-aghayan (changes requested):

test/query_component_test.dart
hanj7221: The default value is now set to byAdvertiser.

test/query_component_test.dart
hanj7221: execQuery depends on google js api to make requests. I tried and it seems like the js wrapper functions cannot be called in testing environment? We could talk more about this.

test/query_component_test.dart
ruben-aghayan: should add test for that

test/query_component_test.dart
```diff
@@ -58,156 +59,333 @@ class OverrideDirective {}
...
+          generateInsertionOrder(
+              advertiserId, mediaPlanIdA, ioIdA, startDate, endDate, spendingA)
+        ];
+      });
+
+      test('highlight underpacing unchecked invokes populate(single-IO, false)',
+          () async {
+        await queryComponentPO.clickPopulate();
+
+        await Future.delayed(Duration(seconds: 2));
```
ruben-aghayan, line 381: drop a comment by these futures to indicate why they are here. normally we try to avoid these (even though theyre necessary here) so an explanation may be necessary

test/testing/query_component_radio_button_po.dart
```diff
@@ -0,0 +1,26 @@
+import 'package:pageloader/pageloader.dart';
```
ruben-aghayan: opt: recommend putting this in the same file as an inner class so the po files track the component files

ruben-aghayan (approved):
looks good some small comments

PR #66, opened 2020-07-23
Query page basic ui

### LINKED ISSUE #3: Query page UI [1~2 day]
Query page should display the predetermined advertiser id and the insertion order id. This will be changed later to display input box to allow for user interaction.

The query page should also include a submit-query button.

Cost estimate: 1~2 day

### LINKED ISSUE #8: Update query page UI with user input advertiser id [ < 1 day]
Update the query page to allow users input an advertiser id that is within their credential access right.

Cost estimate: < 1 day

### LINKED ISSUE #12: Update query page UI with user input advertiser id & media plan id [1 day]
In addition to advertiser id, also allow users to enter a media plan id.

Cost estimate: 1 day

hanj7221:
Closes #3.

Query page UI:
Screen Shot 2020-07-23 at 3 31 00 PM

Display error when user enters non-number into the id fields:
Screen Shot 2020-07-23 at 3 31 20 PM

Display error when user doesn't enter the required parameters:
Screen Shot 2020-07-23 at 3 46 36 PM

Populate button has a spinner that indicates request is still in progress:
Screen Shot 2020-07-23 at 3 31 51 PM

### DIFF
--- a/lib/root_component.html
+++ b/lib/root_component.html
@@ -1,3 +1,9 @@
-<h2 debugId="title">{{title}}</h2>
+<nav class="title-row shadow-sm navbar navbar-light bg-white">
+    <a class="navbar-brand" href="#" id="title" debugId="title">
+        <img src="https://developers.google.com/ads/images/logo_display_video_360_192px.svg"
+             width="30" height="30" class="d-inline-block align-top" alt="">
+        {{title}}
+    </a>
+</nav>
 <query></query>
 <credential></credential>
--- a/lib/root_component.html
+++ b/lib/root_component.html
@@ -1,7 +1,6 @@
 <nav class="title-row shadow-sm navbar navbar-light bg-white">
     <a class="navbar-brand" href="#" id="title" debugId="title">
-        <img src="https://developers.google.com/ads/images/logo_display_video_360_192px.svg"
-             width="30" height="30" class="d-inline-block align-top" alt="">
+        <img src="assets/logo_display_video.svg" width="30" height="30" class="d-inline-block align-top" alt="">
         {{title}}
     </a>
 </nav>
--- a/lib/src/query_component.dart
+++ b/lib/src/query_component.dart
@@ -22,47 +22,69 @@ import 'util.dart';
   directives: [bsAccordionDirectives, BsInput, coreDirectives, formDirectives],
 )
 class QueryComponent implements OnInit {
-  // Names used in css.
+  // Values used in html.
+  static const queryTypeChoices = QueryType.values;
   static const requestSectionTitle = 'Request Parameter';
   static const advertiserParameterName = 'advertiserId';
   static const mediaPlanParameterName = 'campaignId';
-  static const numberOnlyPattern = '^[0-9]*\$';
   static const insertionOrderParameterName = 'insertionOrderId';
+  static const numberOnlyPattern = '^[0-9]+\$';
   static const underpacingCheckBoxName =
       'Highlight underpacing insertion orders';
   static const populateButtonName = 'populate';
-  static const parameterAlertText = 'Missing required parameter values.';
-  static bool showAlert = false;
-  static bool showSpinner = false;
+
+  // Stores accordion panel selection, default is byAdvertiser.
+  QueryType queryType = QueryType.byAdvertiser;
+
+  // Stores input box values.
+  String advertiserId;
+  String mediaPlanId;
+  String insertionOrderId;
+
+  // Stores checkbox value, default is false.
+  bool highlightUnderpacing = false;
+
+  // Controls when to show the spinner on the populate button.
+  bool showSpinner = false;
 
   final QueryService _queryService;
   final ExcelDart _excel;
-
-  // Default selection is byAdvertiser.
-  static QueryType queryType = QueryType.byAdvertiser;
-  static List<QueryType> queryTypeChoices = QueryType.values;
-
-  static String advertiserId;
-  static String mediaPlanId;
-  static String insertionOrderId;
-
-  static bool highlightUnderpacing = false;
 
   QueryComponent(this._queryService, this._excel);
 
   @override
   void ngOnInit() async => await _excel.loadOffice();
 
-  void onClick() async {
-    // Determines if all required parameters are filled in.
-    if (_missingRequiredParameters()) {
-      showAlert = true;
-      showSpinner = false;
-      return;
-    }
+  /// Disables the populate button if missing required input ids or
+  /// if the ids are not integers.
+  ///
+  /// Binds to the [disabled] attribute on the populate button.
+  bool disablePopulateButton() {
+    var invalid = false;
+    final numberOnly = RegExp(numberOnlyPattern);
+    switch (queryType) {
+      case QueryType.byMediaPlan:
+        invalid = mediaPlanId == null || !numberOnly.hasMatch(mediaPlanId);
+        continue checkAdvertiser;
 
-    // All required parameters are filled in, show spinner and hide alert.
-    showAlert = false;
+      case QueryType.byInsertionOrder:
+        invalid =
+            insertionOrderId == null || !numberOnly.hasMatch(insertionOrderId);
+        continue checkAdvertiser;
+
+      checkAdvertiser:
+      case QueryType.byAdvertiser:
+        return advertiserId == null ||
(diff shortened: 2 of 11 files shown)


pubspec.yaml
```diff
@@ -16,6 +17,7 @@ dependencies:
   fixnum: ^0.10.11
   js: ^0.6.1+1
   intl: ^0.15.8
+  ng_bootstrap: any
```
hanj7221, line 20: Found a material design package that works, but had to downgrade angular and angular forms

arulprakash-g (approved):

lib/root_component.html
```diff
@@ -1,3 +1,9 @@
-<h2 debugId="title">{{title}}</h2>
+<nav class="title-row shadow-sm navbar navbar-light bg-white">
+    <a class="navbar-brand" href="#" id="title" debugId="title">
+        <img src="https://developers.google.com/ads/images/logo_display_video_360_192px.svg"
```
ruben-aghayan: can we provide the image as a local resource?

lib/src/query_component.dart
```diff
@@ -18,31 +19,51 @@ import 'util.dart';
...
+  static bool showAlert = false;
+  static bool showSpinner = false;
+
   final QueryService _queryService;
   final ExcelDart _excel;
 
   // Default selection is byAdvertiser.
-  QueryType queryType = QueryType.byAdvertiser;
-  List<QueryType> queryTypeChoices = QueryType.values;
+  static QueryType queryType = QueryType.byAdvertiser;
```
ruben-aghayan: actually think this is more appropriate as an instance type. this is the currently selected query type right?

lib/src/query_component.dart
```diff
@@ -83,6 +107,25 @@ class QueryComponent implements OnInit {
   /// Used in [query_component.html] as part of the radio button id.
   String getQueryTypeShortName(QueryType queryType) => queryType.shortName;
 
+  bool _missingRequiredParameters() {
+    switch (queryType) {
+      case QueryType.byAdvertiser:
```
ruben-aghayan: there is shared logic here, can we make use of fall through?

lib/src/query_component.dart
```diff
@@ -83,6 +107,25 @@ class QueryComponent implements OnInit {
...
+        return advertiserId == null || advertiserId.isEmpty;
+
+      case QueryType.byMediaPlan:
```
ruben-aghayan: can i click even when i have alphabet text entered? the alert will show but is there anything stopping from the request being made?

lib/src/query_component.html
```diff
@@ -1,17 +1,81 @@
...
+        <bs-accordion [closeOthers]="true">
+            <bs-accordion-panel class='query-type-name'
+                                id="{{getQueryTypeShortName(choice)}}-panel"
+                                *ngFor="let choice of queryTypeChoices"
+                                heading="{{getQueryTypeName(choice)}}"
+                                [isOpen]="queryType===choice"
+                                (click)="queryType=choice">
+                <div class="section-name">{{requestSectionTitle}}</div>
+                <form class="form-group">
+                    <div>
```
ruben-aghayan: why doesnt this need an ngif like the other two

web/main.scss
```diff
@@ -0,0 +1,55 @@
...
+.underpacing-checkbox-name {
+  font: 16px Roboto, "Helvetica Neue", sans-serif;
+}
+
+.btn.btn-success {
+  text-transform: uppercase;
+  font: 13px Roboto, sans-serif;;
+}
+
+.btn.btn-primary{
```
ruben-aghayan, line 59: try to use one font family for everything

web/main.scss
```diff
@@ -0,0 +1,55 @@
...
+.parameter-input {
+  color: $regular-font-color;
+  font: 14px "Roboto Mono", monospace;
+}
+
+.underpacing-checkbox {
+  margin-top: 10px;
+  margin-bottom: 10px;;
+}
+.underpacing-checkbox-name {
```
ruben-aghayan: you can put this inside .underpacing-checkbox

ruben-aghayan (changes requested):
any tests? there is some non-css logic like the spinner and input validation. also showing the inputs fields mutually exclusively.

lib/src/query_component.dart
hanj7221: Ohh good catch, maybe I should disable the button until all inputs are filled in & valid. Seems like the error message should be used when I get to error handling for failed query request.

lib/src/query_component.html
hanj7221: Since advertiser id is required no matter what query type it is.

web/main.scss
```diff
@@ -1,55 +1,62 @@
 $title-section-background-color: #202124;
 $body-background-color: #E9ECEF;
 
+$title-font: "Google Sans";
```
hanj7221, line 4: So three font families: one for the title, one for the body, and one for the parameter input box.

web/main.scss
hanj7221: see [here](https://github.com/googleinterns/dv360-excel-plugin/pull/66#discussion_r460973727)

hanj7221 (commented):
How should I test for the spinner? It is removed(ngif) by default, and only when the query still in the processing stage it is added back.

lib/src/query_component.dart
```diff
@@ -22,47 +22,69 @@ import 'util.dart';
...
+  /// if the ids are not integers.
+  ///
+  /// Binds to the [disabled] attribute on the populate button.
+  bool disablePopulateButton() {
+    var invalid = false;
+    final numberOnly = RegExp(numberOnlyPattern);
+    switch (queryType) {
+      case QueryType.byMediaPlan:
+        invalid = mediaPlanId == null || !numberOnly.hasMatch(mediaPlanId);
+        continue checkAdvertiser;
```
ruben-aghayan, line 68: woah thats cool

does io not need media plan id?

test/query_component_test.dart
```diff
@@ -60,9 +61,9 @@ class QueryTestComponent {}
 void main() {
   group('In $QueryComponent,', () {
     const emptyEntry = '';
-    const advertiserId = 'advertiser-id';
-    const mediaPlanIdA = 'media-plan-id-a';
-    const ioIdA = 'io-id-a';
+    const advertiserId = '10000';
+    const mediaPlanIdA = '20000';
```
ruben-aghayan: opt: we typically use number e.g. mediaPlanId1

test/query_component_test.dart
```diff
@@ -182,33 +183,37 @@ void main() {
...
     });
 
-    group('selecting ByAdvertiser with', () {
-      const mediaPlanIdB = 'media-plan-id-b';
-      const ioIdB = 'io-id-b';
-      const ioIdC = 'io-id-c';
-      const ioIdD = 'io-id-d';
+    group('selecting ByAdvertiser panel with', () {
+      const mediaPlanIdB = '20001';
+      const ioIdB = '30001';
```
ruben-aghayan: instantiate these up with the other test consts

test/query_component_test.dart
```diff
@@ -249,18 +254,32 @@ void main() {
         ];
       });
 
-      test(
-          'highlight underpacing unchecked invokes populate(multiple-IO, false)',
+      tearDown(() => clearInteractions(mockQueryService));
```
ruben-aghayan, line 303: this shouldnt be necessary if you are making a new instance for every test

test/query_component_test.dart
```diff
@@ -11,6 +11,7 @@ import 'package:dv360_excel_plugin/src/query_service.dart';
 import 'package:dv360_excel_plugin/src/util.dart';
```
ruben-aghayan, line 11: should test the spinner somewhere

lib/src/query_component.dart
hanj7221: haha nope :)

test/query_component_test.dart
hanj7221: Since i check for the input format, having non-integer ids will actually disable the button

ruben-aghayan (approved):

test/query_component_test.dart
hanj7221: I'm using singleton pattern (needs component provider overwrite) here so it always returns the same instance.

PR #67, opened 2020-07-24
Create excel sheet and table with unique names and improve formats

### LINKED ISSUE #36: Excel spreadsheet needs to be created with unique names

hanj7221:
Closes #36 

Sample spreadsheet format:
Screen Shot 2020-07-23 at 10 44 32 PM


lib/src/excel.dart
```diff
@@ -35,49 +34,65 @@ class ExcelDart {
   static const _resultTableColumnStartAddress = 'A';
   static const _resultTableRowStartAddress = 4;
 
-  static const _fontName = 'Roboto';
```
hanj7221, line 38: Default font(Calibri) looks better

lib/src/excel.dart
```diff
@@ -164,46 +185,49 @@ class ExcelDart {
...
 
     // Waits for population to be complete.
     await populateCompleter.future;
 
     // Conditionally format the budget column once the table is set.
-    await _formatBudgetColumn(tableName);
+    await _formatBudgetAndSpentColumns(tableName);
 
     // Highlight underpacing insertion orders if highlightUnderpacing is true.
-    if (highlightUnderpacing) await _highlightUnderpacingRows(tableName);
```
hanj7221, line 179: Always apply the conditional formatting formula even if `highlightUnderpacing` is false.

I changed the formula to [this](https://github.com/googleinterns/dv360-excel-plugin/blob/9d4a03356e5c76bf6d3aae5bb8ed06ad7e6706ff/lib/src/excel.dart#L250-L251) so that users can also change the true/false value in the cell to highlight/no highlight.

lib/src/excel.dart
```diff
@@ -35,49 +34,65 @@ class ExcelDart {
...
+  static const _budgetUnitColumnName = 'Budget_Unit';
+  static const _budgetColumnName = 'Budget_Amount';
+  static const _spentColumnName = 'Spent_Amount';
+  static const _currentPacingColumnName = 'Current_Pacing_%';
+  static const _pacingPeriodColumnName = 'Pacing_Period';
+  static const _pacingTypeColumnName = 'Pacing_Type';
+  static const _startDateColumnName = 'Start_Date';
+  static const _endDateColumnName = 'End_Date';
+  static const _dailyMaxColumnName = 'Daily_Max';
+  static const _dailyMaxImpressionColumnName = 'Daily_Max_Impression';
```
arulprakash-g, line 68: Do you need to split this into 2 cols - you have budget unit. Aim to be consistent ie either split budget amount col into 2 by type or merge these 2 into 1 col. I prefer the later.

arulprakash-g (approved):

ruben-aghayan (approved):
looks good. the display name column is really wide though, do you control its width?

Jin: ohhh I replaced all advertiser names with 'name' when I took the screenshot. It should auto-fit to the content, it's wide because of some long advertiser names.

lib/src/excel.dart
hanj7221: Sounds good, I will address this in a separate pull request since I'll need to change the proto and some other tests.

New issue created: #68

PR #72, opened 2020-07-25
Create a protocol buffer to represent a rule

### LINKED ISSUE #69: Create protocol buffer to represent rules

thu5:
This PR creates a protobuf to represent a rule, action, schedule. Resolves #69.

### DIFF
--- a/rule.proto
+++ b/rule.proto
@@ -1,0 +1,221 @@
+syntax = "proto2";
+package dv360_excel_plugin;
+
+/* A message that represents a scheduled rule for manipulating DV360 entities.
+ */
+message Rule {
+  /* required
+   * The name of the rule, set by the user.
+   */
+  optional string name = 1;
+
+  /* required
+   * The id that uniquely identified each rule.
+   */
+  optional int32 id = 2;
+
+  // required
+  optional Action action = 3;
+
+  // required
+  optional Condition condition = 4;
+
+  // required
+  optional Schedule schedule = 5;
+}
+
+/*
+ * A message representing the action applied to a DV360 entity.
+ */
+message Action {
+  enum Type {
+    UNSPECIFIED_TYPE = 0;
+    CHANGE_LINE_ITEM_STATUS = 1;
+    CHANGE_BIDDING_STRATEGY = 2;
+    CHANGE_FIXED_BID_PRICE = 3;
+    DUPLICATE_LINE_ITEM = 4;
+  }
+
+  /* required
+   * The type of action.
+   */
+  optional Type type = 1;
+
+  /*
+   * The action parameters to change line items' status.
+   */
+  message ChangeLineItemStatusParams {
+    // Line items to which the action applies.
+    repeated string line_items = 1;
+
+    enum Status {
+      UNSPECIFIED_STATUS = 0;
+      ACTIVE = 1;
+      PAUSED = 2;
+    }
+
+    /* required
+     * The status of which the line item will be changed to.
+     * Can be set to either active or paused.
+     */
+    optional Status status = 2;
+  }
+
+  /*
+   * The action parameters to change line items' bidding strategy.
+   */
+  message ChangeBiddingStrategyParams {
+    // Line items to which the action applies.
+    repeated string line_items = 1;
+
+    enum Strategy {
+      UNSPECIFIED_STRATEGY = 0;
+      FIXED_BID = 1;
+      MAXIMIZE_SPEND_BID = 2;
+      PERFORMANCE_GOAL_BID = 3;
+    }
+
+    /* required
+     * The bidding strategy of which the line item will be changed to.
+     * Can be FixedBid, MaximizeSpendBid, or PerformanceGoalBid.
+     */
+    optional Strategy strategy = 2;
+  }
+
+  /*
+   * The action parameters to change line items' FixedBid price.
+   */
+  message ChangeFixedBidPriceParams {
+    // Line items to which the action applies.
+    repeated string line_items = 1;
+
+    enum Change {
+      UNSPECIFIED_CHANGE = 0;
+      INCREASE = 1;
+      DECREASE = 2;
+    }
+
+    /* required
+     * Whether the FixedBid price is increased or decreased.
+     */
+    optional Change change = 2;
+
+    /* required
+     * How much the FixedBid price is changed.
+     * Can either be absolute value or percentage.
+     */
+    oneof quantity {
+      int32 value = 3;
+      int32 percentage = 4;
+    }
+  }
+
+  /*
+   * The action parameters to duplicate line items.
+   */
+  message DuplicateLineItemParams {
+    // Line items to which the action applies.
+    repeated string line_items = 1;
+
+    /* required
+     * The insertion order at which the line items will be duplicated.
+     */
+    optional string insertion_order_destination = 2;
+  }
+
+  /* required
+   * The parameters used for each type of action.
+   */
+  oneof params {
+    ChangeLineItemStatusParams change_line_item_status_params = 2;
+    ChangeBiddingStrategyParams change_bidding_strategy_params = 3;
+    ChangeFixedBidPriceParams change_fixed_bid_price_params = 4;
+    DuplicateLineItemParams duplicate_line_items_params = 5;
+  }
+}
+
+/* required
+ * A message representing the condition that will be checked before the action is executed.
+ */
+message Condition {
+  enum Type {
+    UNSPECIFIED_TYPE = 0;
+    CPM = 1;
+    CPA = 2;
+    CPC = 3;
+    SPENDING = 4;
+  }
+
+  /* required
+   * The type of condition.
+   * Can be dependent on the cost-per-mille (CPM), cost-per-action (CPA), or cost-per-click (CPC),
+   * or spending of the line item.
+   */
+  optional Type type = 1;
+
+  enum Relation {
+    UNSPECIFIED_RELATION = 0;
+    LESSER = 1;
+    GREATER = 2;
+  }
+
+  /* required
+   * The relation with which the CPM, CPC, CPA, or spending will be compared.
+   * Can be either lesser than or greater than.
+   */
+  optional Relation relation = 2;
+
+  /* required
+   * The value which the CPA, CPC, CPA or spending will be compared against.
+   */
+  optional int32 value = 3;
+}
+
+/* required
+ * A message representing the schedule the rule will be run.
+ */
+message Schedule {
+  enum Type {
+    UNSPECIFIED_SCHEDULE = 0;
+    REPEATING = 1;
+    ONE_TIME = 2;
+  }
+
+  /* required
+   * The rule can be scheduled for either repeating or one-time.
+   */
+  optional Type type = 1;
+
+  /*
+   * The parameters for a repeating rule.
+   */
+  message RepeatingParams {
+    /* required
+     * The cron expression representing the repeating schedule.
+     */
+    optional string cron_expression = 1;
+  }
+
+  /*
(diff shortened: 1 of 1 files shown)


rule.proto
```diff
@@ -0,0 +1,221 @@
...
+  /* required
+   * The id that uniquely identified each rule.
+   */
+  optional int32 id = 2;
+
+  // required
+  optional Action action = 3;
+
+  // required
+  optional Condition condition = 4;
```
thu5: I changed the design so that each rule only has one optional condition.

rule.proto
```diff
@@ -0,0 +1,221 @@
...
+   * The type of action.
+   */
+  optional Type type = 1;
+
+  /*
+   * The action parameters to change line items' status.
+   */
+  message ChangeLineItemStatusParams {
+    // Line items to which the action applies.
+    repeated string line_items = 1;
```
ruben-aghayan: why string vs int

rule.proto
```diff
@@ -0,0 +1,221 @@
...
+     * The status of which the line item will be changed to.
+     * Can be set to either active or paused.
+     */
+    optional Status status = 2;
+  }
+
+  /*
+   * The action parameters to change line items' bidding strategy.
+   */
+  message ChangeBiddingStrategyParams {
```
ruben-aghayan: if you are not implementing this yet, i would hold off on making the proto for it

rule.proto
thu5: The DV360 API returns the [lineItemId](https://developers.google.com/display-video/api/reference/rest/v1/advertisers.lineItems#resource:-lineitem) as a string in int64 format, so I just chose string. Should I change it to int64?

rule.proto
thu5: Ok, that makes sense!

rule.proto
hanj7221: I remembered I had the exact same discussion earlier with my design as well. My justification was that I'm displaying this as a string in Excel, so I kept the string type in the proto. I would actually recommend keeping this as int64, so that when you parse the LI ids from input box, it will throw error if the string ids contain non numbers.

rule.proto
hanj7221: You can remove the optional comment if the field is optional. The idea is use 
```
// required
optional field = 1
```
to replace 
```
required field = 1
```
Here is the [comment](https://github.com/DV360-spreadsheet-plugin/dv360-excel-plugin/pull/30#discussion_r444429947) i received from Ruben regarding this.

rule.proto
```diff
@@ -17,7 +17,7 @@ message Rule {
   // required
   optional Action action = 3;
 
-  // required
+  // optional
```
hanj7221: You can remove the optional comment if the field is optional. The idea is use 
```
// required
optional field = 1
```
to replace 
```
required field = 1
```
Here is the comment i received from Ruben regarding this:
> required has fallen out of favor b/c its not backwards compatible. consider optional instead

thu5: Oh gotcha. Ruben pointed out that for this first iteration, I'm not implementing the conditions, so I just removed it altogether.

rule.proto
thu5: That makes sense!

rule.proto
```diff
@@ -0,0 +1,95 @@
+syntax = "proto2";
+package dv360_excel_plugin;
+
+/* A message that represents a scheduled rule for manipulating DV360 entities.
```
cathyang-work: If comment is only one line, can we make it like this?
`/* A message that represents a scheduled rule for manipulating DV360 entities. */`

Also, if the comment is inside a message, maybe we can just use //. In fact in dv3 public api proto files, it only use `// ...`, there is no `/* ... */`.

e.g. 
```
// required
// The name of the rule, set by the user.
```

https://developers.google.com/protocol-buffers/docs/proto3#adding_comments

rule.proto
```diff
@@ -0,0 +1,95 @@
...
+     * The cron expression representing the repeating schedule.
+     */
+    optional string cron_expression = 1;
+  }
+
+  /* required
+   * The parameters used for each type of scheduling.
+   */
+  oneof params {
+    RepeatingParams repeating_params = 2;
```
cathyang-work, line 80: why this is a 2 instead of 1?

rule.proto
```diff
@@ -0,0 +1,95 @@
...
+
+  /* required
+   * The type of action.
+   */
+  optional Type type = 1;
+
+  /*
+   * The action parameters to change line items' status.
+   */
+  message ChangeLineItemStatusParams {
```
cathyang-work: Can we put the params proto outside of the Action proto? If we put params protos inside  Action proto, the Action proto will grow too big when we add more params protos.

rule.proto
```diff
@@ -0,0 +1,95 @@
...
+    // Line items to which the action applies.
+    repeated int64 line_items = 1;
```
cathyang-work: line_item_ids? and it is required as well?

rule.proto
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  /* required
+   * The parameters used for each type of action.
+   */
+  oneof params {
+    ChangeLineItemStatusParams change_line_item_status_params = 2;
+  }
+}
+/* required
```
cathyang-work: extra line

cathyang-work (changes requested):
Are we missing the Condition proto that you mentioned in the design doc?

rule.proto
thu5: Ok, sure!

thu5:
> Are we missing the Condition proto that you mentioned in the design doc?

I'm not implementing the condition right now, so I followed Ruben's suggestion and removed it for now.

rule.proto
thu5: Field number 1 is used by ['type'](https://github.com/googleinterns/dv360-excel-plugin/blob/8aa67c43e9e073d0f2a8fa3b5cd0538eaedbe459/rule.proto#L77).

cathyang-work:
> > Are we missing the Condition proto that you mentioned in the design doc?
> 
> I'm not implementing the condition right now, so I followed Ruben's suggestion and removed it for now.

I think Ruben is referring to making more params (e.g. ChangeBiddingStrategyParams  proto). In the file, it already include one param proto (ChangeLineItemStatusParams). Without Condition, how does the code decide whether to trigger the action or not. But I guess you can submit another request when you realize you need it. This is not a submission stopper for me, more like a reminder.

rule.proto
thu5: Ok sure!

rule.proto
thu5: Oops, will change it

rule.proto
```diff
@@ -0,0 +1,81 @@
...
+    PAUSED = 2;
+  }
+
+  // required
+  // The status of which the line item will be changed to.
+  // Can be set to either active or paused.
+  optional Status status = 2;
+}
+
+// required
```
ruben-aghayan: does required do anything here

ruben-aghayan (approved):

rule.proto
```diff
@@ -0,0 +1,81 @@
...
+
+  enum Status {
+    UNSPECIFIED_STATUS = 0;
+    ACTIVE = 1;
+    PAUSED = 2;
+  }
+
+  // required
+  // The status of which the line item will be changed to.
+  // Can be set to either active or paused.
```
cathyang-work: Is it necessary if the Enum above already shows that?

rule.proto
thu5: Oops, no it doesn't.

rule.proto
thu5: Sorry, which Enum?

rule.proto
```diff
@@ -0,0 +1,221 @@
+syntax = "proto2";
```
gg-qinp, line 1: why not use proto3

rule.proto
```diff
@@ -0,0 +1,221 @@
...
+
+/* A message that represents a scheduled rule for manipulating DV360 entities.
+ */
+message Rule {
+  /* required
+   * The name of the rule, set by the user.
+   */
+  optional string name = 1;
+
+  /* required
```
gg-qinp: this is required for get but should be generated by server when creating, correct?

rule.proto
gg-qinp: +1 for int, our internal implementation was using int64, the returned json converted it to string.

rule.proto
gg-qinp: +1 for int64, the returned json converted int64 into string. internally we use int64

rule.proto
```diff
@@ -0,0 +1,221 @@
...
+   * The type of action.
+   */
+  optional Type type = 1;
+
+  /*
+   * The action parameters to change line items' status.
+   */
+  message ChangeLineItemStatusParams {
```
gg-qinp: probably you need advertiser id as well. check the API user guide, the request requires advertiser id.

rule.proto
cathyang-work: Status

cathyang-work (approved):

rule.proto
```diff
@@ -0,0 +1,81 @@
...
+
+  // required
+  // The rule can be scheduled for either repeating or one-time.
+  optional Type type = 1;
+
+  // The parameters for a repeating rule.
+  message RepeatingParams {
+    // required
+    // The cron expression representing the repeating schedule.
+    optional string cron_expression = 1;
```
gg-qinp, line 74: is this directly consumed by Scheduler? If the format is like the cron job definition in Linux '* * * * * *` and input by user, seems it is not user friendly.

rule.proto
thu5: Oh gotcha, I'll remove it.

rule.proto
thu5: Yes, Scheduler takes cron expressions. I was thinking the UI could look something like this and just transform it into a cron expression on the client side:

Screen Shot 2020-07-29 at 3 18 23 PM

Should I separate it out to individual fields?

rule.proto
thu5: The proto Jin created uses proto2, so I thought I'd keep it consistent. I don't really use any features that proto2 has that proto3 doesn't, so should I change it?

rule.proto
thu5: Yes, I should remove the required comment :)

rule.proto
thu5: Whoops, forgot about it!

rule.proto
```diff
@@ -0,0 +1,81 @@
...
+// A message representing the action applied to a DV360 entity.
+message Action {
+  enum Type {
+    UNSPECIFIED_TYPE = 0;
+    CHANGE_LINE_ITEM_STATUS = 1;
+  }
+
+  // required
+  // The type of action.
+  optional Type type = 1;
```
gg-qinp, line 29: What is Type used for, imaging we need to add a new action increase bidding price, we will add a new Type in Type Enum, add a new entry in oneof params, If it is a 1to1 mapping, seems the Type is useless?

rule.proto
gg-qinp: I think doing conversion in client side works.

rule.proto
thu5: It's not totally redundant. Using the generated classes, having a Type field makes determining the kind of action more readable.

`myrule.action.type == Action_Type.CHANGE_LINE_ITEM_STATUS`
vs
`myrule.action.whichParams() == Action_Params.changeLineItemStatusParams`

Additionally, we can do stuff with the generated Enum.

gg-qinp: ok, make sense from readability perspective.

gg-qinp (approved):

PR #80, opened 2020-07-29
Improving landing page UI

### LINKED ISSUE #2: Landing page UI [< 1 day]
A simple landing page that includes the DV360 logo, title, and brief instructions maybe.

Cost estimate: < 1 day

### LINKED ISSUE #4: Update components when users sign in/out [1 day]
When user is signed in and grants access to the application, the main page should display components for users to select and submit queries.

When user is signed out, the main page should be updated to hide those components and display the sign-on button instead.

Will be completed together with #3.

hanj7221:
Closes #2, Closes #4

Landing page when the app is loaded outside of Excel
Screen Shot 2020-07-28 at 8 25 26 PM

Landing page when the app is loaded inside Excel
Screen Shot 2020-07-28 at 8 25 58 PM

Main page with nav bar on top, allow switching between 'Query Builder' and 'Rules Builder'
Screen Shot 2020-07-28 at 8 28 06 PM


lib/src/credential_component.dart
```diff
@@ -1,24 +0,0 @@
-import 'package:angular/angular.dart';
```
hanj7221, line 1: The credential component is removed. Its functionality (sign in and out) is moved to root component

lib/src/query_component.dart
```diff
@@ -21,7 +21,7 @@ import 'util.dart';
   ],
   directives: [bsAccordionDirectives, BsInput, coreDirectives, formDirectives],
 )
-class QueryComponent implements OnInit {
```
hanj7221, line 24: Loading the Office JS library on component init is also moved to root component, since it needs to know whether the app is sideloaded in Excel or not to determine what content to render.

test/root_component_test.dart
```diff
@@ -0,0 +1,75 @@
+@TestOn('browser')
```
hanj7221, line 1: These tests are not comprehensive. Additional tests will be added in the next commit.

lib/root_component.dart
```diff
@@ -1,13 +1,45 @@
...
+class RootComponent implements OnInit {
+  // Names used in html.
   final title = 'Display & Video 360';
+  final welcomeTitle = 'Welcome!';
+  final welcomeMessage =
+      'Discover what DV360 Excel Add-in can do for you today';
+  final sideloadMessage = 'Please sideload the add-in to continue';
+  final queryBuilderTabName = 'Query Builder';
+  final rulesBuilderTabName = 'Rules Builder';
+  final signOnButtonName = 'GET STARTED';
```
ruben-aghayan: opt: all caps effect can be done using css.

lib/root_component.dart
```diff
@@ -1,13 +1,45 @@
...
+  final signOutButtonName = 'Sign out';
+
+  // The current sign-in status.
+  bool isUserValidated;
```
ruben-aghayan: initialize this as false so you arent checking thats its null in the template

lib/root_component.dart
```diff
@@ -1,13 +1,45 @@
...
+
+  // The current environment status.
+  bool isExcelEnvironment;
```
ruben-aghayan: consider making these bools private and having "shouldShowSignIn" and "shouldShowQuery" bools instead. that way the logic deciding whether to show something is in the dart and not the template.

lib/root_component.dart
```diff
@@ -1,13 +1,45 @@
...
+  final sideloadMessage = 'Please sideload the add-in to continue';
+  final queryBuilderTabName = 'Query Builder';
+  final rulesBuilderTabName = 'Rules Builder';
+  final signOnButtonName = 'GET STARTED';
+  final signOutButtonName = 'Sign out';
+
+  // The current sign-in status.
+  bool isUserValidated;
+
+  // The current environment status.
```
ruben-aghayan, line 29: im missing where user validated and excelEnvironment are set.

lib/src/credential_service.dart
```diff
@@ -49,29 +61,25 @@ class CredentialService {
         discoveryDocs: [_dv3DiscoveryUrl, _dbmDiscoveryUrl],
         scope: _scope);
 
-    await GoogleAPI.client.init(initArgs).then(allowInterop((value) {
+    final completer = Completer<bool>();
```
ruben-aghayan, line 64: just a thought, is it possible to move the completer logic into your wrapper class? would be nice to abstract that detail away from the caller.

lib/src/office_js.dart
```diff
@@ -18,155 +18,14 @@ import 'package:js/js.dart';
...
-
-  /// The runtime environment of the add-in.
-  external static Context get context;
-
-  /// The result of an asynchronous call.
-  external static AsyncResultStatus get asyncResultStatus;
 }
 
 /// Input argument to [onReady()].
 @JS()
```
ruben-aghayan, line 24: what happened do you not need this class anymore?

ruben-aghayan (changes requested):
wow this look great!!

lib/root_component.dart
hanj7221: so i guess it's ok that user sees the landing page first and once the initial sign-in state is returned from the future, it then jumps to the main page

lib/root_component.dart
hanj7221: [here](https://github.com/googleinterns/dv360-excel-plugin/pull/80/files/17ba7638638ab0471dc29e4f421b110acecce154#diff-0ddd5d14fdbdacb8aa08fb8299ded715R38-R44)

lib/src/office_js.dart
hanj7221: I added those wrapper functions when i thought i had to use the office dialog box (long long time ago). This class is not used very often, so I totally forgot to delete them :(

lib/src/credential_service.dart
hanj7221: Good point! I will address this in a separate PR (issue created #81)

lib/root_component.dart
ruben-aghayan: aha there it is

test/root_component_test.dart
```diff
@@ -85,6 +85,9 @@ void main() {
     test(
         'handleClientLoad(), initClient() and loadOffice()'
         'are invoked during init', () async {
+      when(mockCredential.initClient()).thenAnswer((_) => Future.value(true));
```
hanj7221, line 88: These extra setups are to ensure that [_updatePage()](https://github.com/googleinterns/dv360-excel-plugin/pull/80/commits/a4f2c4b74f028fa998a59a1b9fc6d12e26a49a70#diff-0ddd5d14fdbdacb8aa08fb8299ded715R48) won't throw when the bools are null.

hanj7221: These extra setups are to ensure that [_updatePage()](https://github.com/googleinterns/dv360-excel-plugin/pull/80/commits/a4f2c4b74f028fa998a59a1b9fc6d12e26a49a70#diff-0ddd5d14fdbdacb8aa08fb8299ded715R48) won't throw since mock returns 'null' for bools by default

hanj7221 (commented):
Adding showMainPage bool, even if it seems redundant. 
But I guess i want the initial values of showLanding and showMain to both be false, so that the app displays nothing until it figures out the sign-in status etc. 

If i do
```
<div *ngIf = showLandingPage> .... </div>
<div *ngIf =!showLandingPage> .... </div>
```
The main page will always be shown because showLandingPage has default value of false.

lib/root_component.dart
```diff
@@ -1,13 +1,64 @@
...
+  final signOutButtonName = 'Sign out';
+
+  // The current sign-in status.
+  bool _isUserValidated;
+
+  // The current environment status.
+  bool _isExcelEnvironment;
+
+  // Controls when to show landing page.
+  bool showLandingPage = false;
```
ruben-aghayan: make this a getter keyed on excel environment and user validated. same below

ruben-aghayan (approved):

PR #84, opened 2020-08-03
Updating reporting api to query with static start date

hanj7221:


lib/src/query_component.dart
```diff
@@ -93,26 +100,20 @@ class QueryComponent {
...
       // Uses DV360 public APIs to fetch entity data.
-      List<InsertionOrder> insertionOrders =
-          await _queryAndParseInsertionOrderEntityData();
-
-      // Early return if the query returns no insertion orders.
-      if (insertionOrders.isEmpty) return;
-
-      // If query type is not by insertion order, filters out insertion orders
-      // that are not in-flight.
-      insertionOrders = queryType == QueryType.byInsertionOrder
```
hanj7221, line 104: This filter logic is moved to the [public api parser class](https://github.com/googleinterns/dv360-excel-plugin/pull/84/commits/cc16512f8afc9c6ea5415eef7df263df1d227762#diff-3863c7633dfe3db3e4718b1629ac87b5R55-R57).

lib/src/query_component.dart
```diff
@@ -53,6 +53,13 @@ class QueryComponent {
   bool showAlert = false;
   String alertMessage;
 
+  // The earliest startDate that reporting API can take.
+  DateTime reportStartDate = DateTime(
+      DateTime.now().year - 2, DateTime.now().month, DateTime.now().day);
+
```
hanj7221, line 59: Seems like the earliest start date for reporting api is Now() minus 2 years.  I will double check this with Caleb.

lib/src/query_component.dart
```diff
@@ -189,40 +190,33 @@ class QueryComponent {
...
-          .map((io) => Util.convertProtoDateToDateTime(
-              io.budget.activeBudgetSegment.dateRange.startDate))
-          .reduce((date, minDate) => date.isBefore(minDate) ? date : minDate);
-
   void _addSpentToInsertionOrders(List<InsertionOrder> insertionOrders,
       Multimap<String, InsertionOrderDailySpend> spendingMap) {
     for (final io in insertionOrders) {
+      // Continues if the spending map doesn't have spending data for this io
+      // at all.
+      if (!spendingMap.containsKey(io.insertionOrderId)) continue;
```
hanj7221, line 200: This case will happen when an IO has no active spending data and no historical spending data either (the IO is created recently).

lib/src/query_component.dart
```diff
@@ -189,40 +190,33 @@ class QueryComponent {
...
               ? dailySpend.revenue
-              : dailySpend.impression)
-          .reduce((spend, sum) => Util.addStringRevenue(spend, sum));
-
-      io.spent = spent;
+              : dailySpend.impression);
+
+      // Continues if the spending map doesn't have spending data for this io
+      // within [reportStartDate, now].
+      if (activeSpent.isEmpty) continue;
```
hanj7221, line 226: This case will happen when the IO has historical spending data but no active ones.

lib/src/public_api_parser.dart
```diff
@@ -52,6 +52,9 @@ class PublicApiParser {
     } else if (map.containsKey('insertionOrders')) {
       return List.from(map['insertionOrders'])
           .map((ioMap) => _createInsertionOrder(ioMap))
+          .where((io) =>
```
ruben-aghayan, line 55: what does this do?

lib/src/query_component.dart
ruben-aghayan: ok please leave issue with todo
what happens when you submit a time earlier than that?
a const may be better here.

lib/src/query_component.dart
```diff
@@ -93,26 +100,20 @@ class QueryComponent {
...
-      // spendingMap has a [DailySpend] for each insertion order.
-      final spendingMap =
-          await _queryAndParseSpentData(minStartDate, DateTime.now());
+      // impression based) within time window [reportStartDate, Now].
+      Future<Multimap<String, InsertionOrderDailySpend>> reportingApiResponse =
+          _queryAndParseSpentData(reportStartDate, DateTime.now());
+      apiRequestFutures.add(reportingApiResponse);
+
+      // Waits for all requests to finish.
+      final responseList = await Future.wait(apiRequestFutures);
```
ruben-aghayan, line 115: do you have to wait for both to finish to start parsing one or the other?
is the time savings worth the complexity of parsing one before the other finishes?

lib/src/query_component.dart
ruben-aghayan: this is a great comment to leave in the code

lib/src/query_component.dart
hanj7221: Caleb replied that the default look up window is 2 years. If a time earlier than that is provided, the reporting api will respond with an error message in json. So I guess there is not much we can do about it, how about letting the spend field display a message like: 'Spend is not available, please go to the website' when the flight is longer than 2 years?

lib/src/query_component.dart
hanj7221: Sorry if the names are confusing, publicApiResponse and reportingApiResponse here represent parsed response actually. So the future is resolved when both requests respond back and finish parsing. In the way, I think what I'm doing here is also parsing the two responses in two separate isolates (Dart version of threads?) 

I think it keeps a good balance? i.e. on average, public api takes longer to response but less time to parse, reporting api takes less time to response but longer time to parse. 

And then the joining of spending data to IOs happen at the last after the await.

lib/src/query_component.dart
hanj7221: Hmm, couldn't really use const here since DateTime.now() is not const. I guess I will use static final then.

lib/src/public_api_parser.dart
hanj7221: This filters out the IOs that are not in-flight when the query type is byAdvertiser/byCampaign. If the query type is byInsertionOrder, then we don't do any filter. And for byIO, if the requested IO is not in-flight, the budget/spending/start date/end date will be empty.

This logic was [here](https://github.com/googleinterns/dv360-excel-plugin/blob/master/lib/src/query_component.dart#L102-L106) before, and I just moved it to be done during the parsing stage.

ruben-aghayan (approved):

PR #94, opened 2020-08-07
Add service class for Firestore and conversion functions

### LINKED ISSUE #87: Create service class for Firestore

thu5:
This PR adds a wrapper for Firestore and functions to convert between Firestore Documents and protobuf generated Rules. Resolves #87.

Sorry this PR is a little larger. 😅

### DIFF
--- a/rule.proto
+++ b/rule.proto
@@ -79,4 +79,8 @@ message Schedule {
   oneof params {
     RepeatingParams repeating_params = 2;
   }
+
+  // required
+  // The timezone name of the scheduled job from the tz database.
+  optional string timezone = 3;
 }
--- a/rule.proto
+++ b/rule.proto
@@ -82,5 +82,6 @@ message Schedule {
 
   // required
   // The timezone name of the scheduled job from the tz database.
+  // See: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
   optional string timezone = 3;
 }
--- a/server/config.src.yaml
+++ b/server/config.src.yaml
@@ -1,6 +1,11 @@
 clientId: clientid
 clientSecret: clientsecret
+projectId: projectId
+databaseId: (default)
 
 displayVideo360:
   baseURL: https://localhost:8001/
+
+firestore:
+  baseURL: https://localhost:8002/
 
--- a/server/config.yaml
+++ b/server/config.yaml
@@ -1,5 +1,10 @@
 clientId: $CLIENT_ID
 clientSecret: $CLIENT_SECRET
+projectId: $PROJECT_ID
+databaseId: (default)
 
 displayVideo360:
   baseURL: https://displayvideo.googleapis.com/
+
+firestore:
+  baseURL: https://firestore.googleapis.com/
--- a/server/lib/service/firestore.dart
+++ b/server/lib/service/firestore.dart
@@ -44,6 +44,9 @@ class FirestoreClient {
 
   /// Adds a user to the Firestore database with the encrypted refresh token.
   ///
+  /// The [userId] is the `sub` claim of the user's Google ID token.
+  /// See: https://developers.google.com/identity/protocols/oauth2/openid-connect#an-id-tokens-payload
+  ///
   /// Throws an [ApiRequestError] if Firestore API returns an error.
   Future<void> createUser(String userId, String encryptedRefreshToken) async {
     final document = Document()
--- a/server/lib/service/firestore.dart
+++ b/server/lib/service/firestore.dart
@@ -7,10 +7,10 @@ import '../utils.dart';
 /// A class that wraps around Cloud Firestore.
 class FirestoreClient {
   /// The name of collection of users in Firestore.
-  static const usersCollectionName = 'users';
+  static const usersName = 'users';
 
   /// The name of the collection of rules in Firestore.
-  static const rulesCollectionName = 'rules';
+  static const rulesName = 'rules';
 
   /// The name of the encrypted refresh token field in Firestore.
   static const encryptedRefreshTokenFieldName = 'encryptedRefreshToken';
@@ -19,33 +19,30 @@ class FirestoreClient {
   final FirestoreApi _api;
 
   /// The Google Cloud project ID.
-  final String projectId;
+  final String _projectId;
 
   /// The Cloud Firestore database ID.
-  final String databaseId;
+  final String _databaseId;
 
   /// Creates an instance of [FirestoreClient].
   FirestoreClient(
-      Client client, this.projectId, this.databaseId, String baseUrl)
+      Client client, this._projectId, this._databaseId, String baseUrl)
       : _api = FirestoreApi(client, rootUrl: baseUrl);
 
-  /// Adds a rule to the Firestore database.
-  ///
-  /// The [rule] is an instance of [Rule], a protobuf generated class.
+  /// Adds a protobuf generated [Rule] to the Firestore database.
   ///
   /// Throws an [ApiRequestError] if Firestore API returns an error.
   Future<String> createRule(String userId, Rule rule) async {
     final document = rule.toDocument();
 
     return await _addDocument(
-      '/$usersCollectionName/$userId',
-      rulesCollectionName,
+      '/$usersName/$userId',
+      rulesName,
       document,
     );
   }
 
-  /// Adds a user to the Firestore database along with the encrypted refresh
-  /// token.
+  /// Adds a user to the Firestore database with the encrypted refresh token.
   ///
   /// Throws an [ApiRequestError] if Firestore API returns an error.
   Future<void> createUser(String userId, String encryptedRefreshToken) async {
@@ -57,7 +54,7 @@ class FirestoreClient {
 
     await _addDocument(
       '',
-      usersCollectionName,
+      usersName,
       document,
       documentId: userId,
     );
@@ -67,7 +64,7 @@ class FirestoreClient {
   ///
   /// Throws an [ApiRequestError] if Firestore API returns an error.
   Future<String> getEncryptedUserRefreshToken(String userId) async {
-    final document = await _getDocument('/$usersCollectionName/$userId');
+    final document = await _getDocument('/$usersName/$userId');
 
     final encryptedRefreshToken =
         document.fields[encryptedRefreshTokenFieldName].stringValue;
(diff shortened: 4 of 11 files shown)


server/lib/service/firestore.dart
```diff
@@ -0,0 +1,117 @@
...
+  final String databaseId;
+
+  /// Creates an instance of [FirestoreClient].
+  FirestoreClient(
+      Client client, this.projectId, this.databaseId, String baseUrl)
+      : _api = FirestoreApi(client, rootUrl: baseUrl);
+
+  /// Adds a rule to the Firestore database.
+  ///
+  /// The [rule] is an instance of [Rule], a protobuf generated class.
```
cathyang-work: Maybe join the two sentences together?

https://dart.dev/guides/language/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph

server/lib/service/firestore.dart
```diff
@@ -0,0 +1,117 @@
+import 'package:googleapis/firestore/v1.dart';
+import 'package:http/http.dart';
+
+import '../proto/rule.pb.dart';
+import '../utils.dart';
+
+/// A class that wraps around Cloud Firestore.
+class FirestoreClient {
+  /// The name of collection of users in Firestore.
+  static const usersCollectionName = 'users';
```
cathyang-work: optional: usersName? Just prefer shorter names if it also works.

server/lib/service/firestore.dart
```diff
@@ -0,0 +1,117 @@
...
+
+  /// The name of the collection of rules in Firestore.
+  static const rulesCollectionName = 'rules';
```
cathyang-work: optional: rulesName? Just prefer shorter names if it also works.

server/lib/service/firestore.dart
```diff
@@ -0,0 +1,117 @@
...
+    final document = rule.toDocument();
+
+    return await _addDocument(
+      '/$usersCollectionName/$userId',
+      rulesCollectionName,
+      document,
+    );
+  }
+
+  /// Adds a user to the Firestore database along with the encrypted refresh
```
cathyang-work: Maybe remove "along" so then it can be in one line?

server/lib/service/firestore.dart
```diff
@@ -0,0 +1,117 @@
...
+  ///
+  /// The [collection] is the name of the collection relative to [parent] where
+  /// the new document will be added.
+  ///
+  /// The [documentId] optionally allows a custom name for the document.
+  ///
+  /// Throws an [ApiRequestError] if Firestore API returns an error.
+  /// Returns the name of the new document.
+  Future<String> _addDocument(
+      String parent, String collection, Document document,
```
cathyang-work: collectionName?

server/lib/service/firestore.dart
```diff
@@ -0,0 +1,117 @@
...
+
+  /// Adds a rule to the Firestore database.
+  ///
+  /// The [rule] is an instance of [Rule], a protobuf generated class.
+  ///
+  /// Throws an [ApiRequestError] if Firestore API returns an error.
+  Future<String> createRule(String userId, Rule rule) async {
+    final document = rule.toDocument();
+
+    return await _addDocument(
```
cathyang-work, line 38: Why this method does not  return the ruleId while the createUser() returns the userId? How do we generate the userId?

server/lib/service/firestore.dart
```diff
@@ -0,0 +1,117 @@
...
+  Future<String> getEncryptedUserRefreshToken(String userId) async {
+    final document = await _getDocument('/$usersCollectionName/$userId');
+
+    final encryptedRefreshToken =
+        document.fields[encryptedRefreshTokenFieldName].stringValue;
+
+    return encryptedRefreshToken;
+  }
+
+  /// Get a [Document] located at [path] from Firestore.
```
cathyang-work: nit: Gets

server/lib/utils.dart
```diff
@@ -0,0 +1,115 @@
...
+  /// This is necessary because [Document.fields] is a map with [Value] values.
+  /// [Value] is a union type. See:
+  /// https://pub.dev/documentation/googleapis/latest/googleapis.firestore.v1/Value-class.html
+  ///
+  /// We will first convert the protobuf [Rule] to a JSON format before
+  /// converting to [Document]. In this way, we can minimize coupling
+  /// between the proto schema and how we construct the document
+  /// (and vice versa).
+  Document toDocument() {
+    // First, transform the Rule into a standard JSON string.
```
cathyang-work: nit: [Rule]

server/lib/utils.dart
```diff
@@ -0,0 +1,115 @@
...
+/// An extension on the [Rule] protobuf generated class.
+extension DocumentConversion on Rule {
+  /// Extends the [Rule] protobuf generated class to be able to convert to a
+  /// Firestore [Document].
+  ///
+  /// This is necessary because [Document.fields] is a map with [Value] values.
+  /// [Value] is a union type. See:
+  /// https://pub.dev/documentation/googleapis/latest/googleapis.firestore.v1/Value-class.html
+  ///
+  /// We will first convert the protobuf [Rule] to a JSON format before
```
cathyang-work: nit: maybe don't need "the protobuf" as it was mentioned above.

server/test/firestore_test.dart
```diff
@@ -0,0 +1,187 @@
...
+    await mockFirestoreServer.close();
+  });
+
+  tearDown(() async {
+    mockFirestoreServer.clear();
+  });
+
+  group('Success case: createRule()', () {
+    setUp(() async {
+      final ruleAsDocument = rule.toDocument();
```
cathyang-work, line 66: optional: ruleDocument?

server/test/utils_test.dart
```diff
@@ -0,0 +1,67 @@
...
+                  'cronExpression': Value()..stringValue = '* * * * *',
+                }))
+          }))
+    };
+
+  group('Rule-Document Conversion:', () {
+    test('toDocument() converts a Rule to an equivalent Document', () async {
+      final ruleAsDocument = rule.toDocument();
+
+      // I haven't overridden [Document.equals()] yet, so currently just
```
cathyang-work: Do you plan to override [Document.equals()]? If so, please put a todo there.

// TODO(@thu5): ...

cathyang-work (changes requested):

server/lib/service/firestore.dart
thu5: Ok, that sounds good.

server/lib/service/firestore.dart
thu5: Yeah, that works!

server/lib/service/firestore.dart
thu5: Nice catch! That makes it clearer.

server/lib/service/firestore.dart
thu5: `createRule()` returns the ruleId because the id for the rule document is generated by Firestore, whereas `createUser()` doesn't because we set its id to the parameter `userId`. 

Should I return the `userId` too? Or just make this clearer in the documentation?

server/test/utils_test.dart
thu5: Okay, sounds good!

rule.proto
```diff
@@ -79,4 +79,8 @@ message Schedule {
   oneof params {
     RepeatingParams repeating_params = 2;
   }
+
+  // required
+  // The timezone name of the scheduled job from the tz database.
```
gg-qinp, line 84: what is `tz` stands for?

server/config.src.yaml
```diff
@@ -1,6 +1,11 @@
 clientId: clientid
 clientSecret: clientsecret
+projectId: projectId
```
gg-qinp, line 3: are these dummy values? then why we need them

server/lib/service/firestore.dart
gg-qinp: I have same question, if we specify userId in client side, what values you plan to use?

server/test/firestore_test.dart
```diff
@@ -0,0 +1,187 @@
...
+      ruleAsDocument.name = ruleName;
```
gg-qinp: does this already set in `rule.toDocument()`?

server/lib/service/firestore.dart
cathyang-work: I think it will be the F1 user id?  It is used in public api request, so user should know their F1 ids? But is it okay to save it in Firestore?

server/config.yaml
```diff
@@ -1,5 +1,10 @@
 clientId: $CLIENT_ID
 clientSecret: $CLIENT_SECRET
+projectId: $PROJECT_ID
```
cathyang-work, line 3: How do you plan to pass in the clientId, clientSecret and projectId?

server/lib/service/firestore.dart
thu5: The userId should be the `sub` field of the user's ID token, which is an obfuscated GAIA ID.

rule.proto
thu5: The tz stands for [time zone](https://en.wikipedia.org/wiki/Tz_database).  I'll add a link to the documentation to make this clearer.

server/config.src.yaml
thu5: Yes, these are just dummy values. `config.src.yaml` is the configuration source file where these values are [used when testing](https://aqueduct.io/docs/application/configure/#configuration-conventions-and-deployment-options), and it's [supposed to have the same structure](https://aqueduct.io/docs/tut/writing-tests/#configuration-template) as `config.yaml`.

server/test/firestore_test.dart
thu5: Oops, that line should be `ruleAsDocument.name = documentName`. I'll fix it 😅

server/config.yaml
thu5: After setting the environment variable either locally or on App Engine, we can pass in the values when the `ServerChannel` is [initialized](https://github.com/googleinterns/dv360-excel-plugin/blob/da668522051694ff1a1579f647130f957cd7d4d1/server/lib/channel.dart#L25).

server/lib/service/firestore.dart
cathyang-work: Can you link the def of the userId token so we can find the `sub` field?

thu5: Sure, here's the [link](https://developers.google.com/identity/protocols/oauth2/openid-connect#an-id-tokens-payload), and I'll add it to the documentation.

cathyang-work (approved):

gg-qinp (approved):

PR #98, opened 2020-08-13
Create models to represent actions and rules

### LINKED ISSUE #88: Create models for rules and actions

thu5:
This PR creates models to represent actions and rules. Additional actions can be added by implementing the Action interface. Resolves #88.


server/lib/model/action.dart
```diff
@@ -0,0 +1,72 @@
...
+      ..changeLineItemStatusParams = (proto.ChangeLineItemStatusParams()
+        ..lineItemIds.addAll(_lineItemIds)
+        ..advertiserId = _advertiserId
+        ..status =
+            proto.ChangeLineItemStatusParams_Status.valueOf(_statusValue));
+  }
+
+  /// Changes the status of the line item(s) using the DV360 client.
+  @override
+  Future<void> run(DisplayVideo360Client client) async {
```
gg-qinp, line 65: Should we record the return and save as running history?

thu5: Yes, I'm planning on implementing running history in another pull request. I'll add a todo?

server/lib/model/rule.dart
```diff
@@ -0,0 +1,20 @@
+import 'dart:async';
+
+import '../proto/rule.pb.dart' as proto;
```
cathyang-work, line 3: Is `../proto/rule.pb.dart` in the codebase yet?

server/lib/model/action.dart
```diff
@@ -0,0 +1,73 @@
...
+
+import '../proto/rule.pb.dart' as proto;
+import '../service/dv360.dart';
+
+/// An interface that represents an action to manipulate DV360 entities.
+abstract class Action {
+  /// Creates an instance of a class that implements [Action], from a proto.
+  ///
+  /// Throws an [UnsupportedError] if the proto's action type is unspecified.
+  factory Action.fromProto(proto.Action action) {
```
cathyang-work: What is the advantage of using factory here? Would creating a concrete method will serve the purpose? We can also use general constructor in other files `ChangeLineItemStatusAction(proto.Action action)` so then our code does not look as complicated.

```
getActionFromProto(proto.Action action) {
   switch (action.type) {
      case proto.Action_Type.CHANGE_LINE_ITEM_STATUS:
        return ChangeLineItemStatusAction(action);
      default:
        throw UnsupportedError(
            '${action.type} is not a supported action type.');
}
```

server/lib/model/rule.dart
thu5: It's generated with protoc, for example with this command:
`protoc --dart_out=server/lib/proto/ rule.proto`

server/lib/model/action.dart
thu5: That makes sense. There's no inherent advantage to using `factory`. 

I got rid of factory, and made `getActionFromProto` a top-level function (instead of a method because otherwise, I'd have to implement it in the subtypes that implement the interface).

cathyang-work: In dart, you can implement a method in abstract class and the subclass inherit the implementation so then you don't need to reimplement in each subclass. With the `Action` in `Action.getActionFromProto(...)`, it is hard to track where method, "getActionFromProto()", comes from in the caller class. 

Search 'concrete methods' in
https://www.learndartprogramming.com/fundamentals/abstract-classes-in-dart/

thu5: I get `error: Missing concrete implementation of 'Action.getActionFromProto'` when I try to create a concrete method.

I think for subclasses that `extends` a class, you can inherit the concrete method, but with subtypes that `implements` an interface, you have to provide an implementation for every method in the interface.

Should I provide a `'noSuchMethod' implementation` for subtypes, [split the interface](https://stackoverflow.com/questions/38750689/declaring-a-method-as-optional-in-abstract-class), or just remove this method and move the logic elsewhere?

thu5: Actually, I guess `factory` does have a use here. Because, if we change this method to a factory, it becomes a constructor that can return subtypes, and since the constructor isn't a part of the interface, it doesn't have to implemented in subtypes.

cathyang-work: I thought you created an abstract class so then the subclasses and `extends` from it. But seems like you use the abstract class as an interface. Yes, if you use it as an interface, you will need to implement all the methods. Have you considered `extends` Action class instead of `implements` it so then you don't need to reimplement the `getActionFromProto()` in every subclass? Or you expect the subclasses will inherit more interfaces? A class can inherit one abstract class and implement multiple interfaces.

thu5: Hm, I wasn't sure if it made sense to allow `changeLineItemStatusAction.getActionFromProto(protoAction)` if I use `extends`.

How about making `getActionFromProto()` a static method in `Action`?

cathyang-work: Sounds good. As we discuss offline, @thu5 will implement a static method in `action.dar`t so the caller class call `Action.getActionFromProto()`

server/lib/model/rule.dart
```diff
@@ -0,0 +1,20 @@
...
+import 'action.dart';
+
+/// A class that represents a rule to manipulate DV360 entities.
+class Rule {
+  /// An [Action] instance that manipulates DV360 entities.
+  final Action _action;
+
+  /// Creates a [Rule] from a [proto.Rule].
+  Rule.fromProto(proto.Rule protoRule)
+      : _action = getActionFromProto(protoRule.action);
```
cathyang-work: Does it need to change to `Action.getActionFromProto()`?

thu5: Completely forgot 😅

cathyang-work (approved):

gg-qinp (approved):

PR #102, opened 2020-08-19
Add user controller

### LINKED ISSUE #89: Create controller for adding users

thu5:
This PR adds a controller for handling operations on users. It also adds utility functions to encrypt/decrypt refresh tokens, and to get the user ID from ID tokens. Closes #89.

### DIFF
--- a/server/config.src.yaml
+++ b/server/config.src.yaml
@@ -4,6 +4,7 @@ projectId: projectid
 databaseId: (default)
 locationId: us-central1
 appEngineService: rules-server
+refreshTokenKey: arefreshtokenkey
 
 displayVideo360:
   baseURL: https://localhost:8001/
--- a/server/config.src.yaml
+++ b/server/config.src.yaml
@@ -4,7 +4,7 @@ projectId: projectid
 databaseId: (default)
 locationId: us-central1
 appEngineService: rules-server
-refreshTokenKey: arefreshtokenkey
+refreshTokenKey: refresh-token-key
 
 displayVideo360:
   baseURL: https://localhost:8001/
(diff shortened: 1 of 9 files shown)


server/config.src.yaml
```diff
@@ -4,6 +4,7 @@ projectId: projectid
 databaseId: (default)
 locationId: us-central1
 appEngineService: rules-server
+refreshTokenKey: arefreshtokenkey
```
cathyang-work: The format should be consistent. and `a` can be ignored. `refresh-token-key?`

server/test/utils_test.dart
```diff
@@ -63,4 +65,32 @@ void main() {
...
+
+      final encrypted = encryptRefreshToken(testRefreshToken, refreshTokenKey);
+      final decrypted = decryptRefreshToken(encrypted, refreshTokenKey);
+
+      expect(decrypted, equals(testRefreshToken));
+    });
+  });
+
+  group('Get user ID from ID token', () {
+    test('encryptRefreshToken() ', () async {
```
gg-qinp: getUserId()

server/lib/controller/user_controller.dart
```diff
@@ -0,0 +1,51 @@
...
+    ContentType('application', 'x-protobuf')
+  ];
+
+  /// Creates a user with a user ID and encrypted refresh token.
+  ///
+  /// Throws an [ArgumentError] if the request does not contain an Authorization
+  /// header. Throws an [ApiRequestError] if the Firestore API returns an error.
+  /// Returns a 200 OK [Response] upon success.
+  @Operation.post()
+  Future<Response> createUser(@Bind.body() List<int> body) async {
```
gg-qinp, line 34: nit:  In general the create() method needs to return the resources created in stead of empty() following the restful design . I assume this function will be called in the background of the server, so it is not customer initiated and visible operation. Also seems that directly binding the body with a message like User has better readability. Maybe we can add a TODO here, and refactor later.

gg-qinp (approved):

server/test/utils_test.dart
thu5: 😅

server/lib/controller/user_controller.dart
thu5: Ok, that makes sense

cathyang-work (approved):

PR #104, opened 2020-08-19
Add run rule controller

### LINKED ISSUE #91: Create controller for running rules

thu5:
This PR adds the controller to handle requests from Cloud Scheduler to run rules. When a job is triggered, it will look up the rule and the user's refresh token in Firestore, create a model, and run the rule. Resolves #91.


server/lib/controller/run_rule_controller.dart
```diff
@@ -0,0 +1,75 @@
...
+
+    // Runs the rule using the DV360 client.
+    await rule.run(dv360);
+
+    return Response.ok({});
+  }
+
+  /// Returns the [Rule] model form of the rule proto.
+  @visibleForTesting
+  Rule getRule(proto.Rule ruleProto) {
```
cathyang-work: Is using `Rule.fromProto(ruleProto);` directly not work in line 51?

server/lib/controller/run_rule_controller.dart
```diff
@@ -0,0 +1,75 @@
...
+
+  /// The base URL of DV360.
+  final String _dv360BaseUrl;
+
+  /// The AES key to encrypt/decrypt refresh tokens.
+  final String _aesKey;
+
+  /// Creates an instance of [RunRuleController].
+  RunRuleController(
+      this._googleApi, this._firestoreClient, this._aesKey, this._dv360BaseUrl);
```
cathyang-work, line 29: Where is `_aesKey` gonna be stored?

server/lib/controller/run_rule_controller.dart
```diff
@@ -0,0 +1,75 @@
...
+    final refreshToken = decryptRefreshToken(encryptedToken, _aesKey);
+
+    // Creates an authenticated client for the user using their refresh token.
+    final userClient = await _googleApi.getUserAccountClient(refreshToken);
+
+    // Creates a DV360 client using the user's authenticated client.
+    final dv360 = DisplayVideo360Client(userClient, _dv360BaseUrl);
+
+    // Runs the rule using the DV360 client.
+    await rule.run(dv360);
```
cathyang-work: What do you think if we have dv360 client to initiate the run instead of passing Dv360 client to the rule. You will need to do some refactoring for the DisplayVideo360Client class. So then in here, we can use:

```
await dv360Client.run(rule);
```

server/lib/controller/run_rule_controller.dart
```diff
@@ -0,0 +1,75 @@
...
+/// A controller that handles running rules.
+class RunRuleController extends ResourceController {
+  /// This provides authenticated clients to interact with Google APIs.
+  final GoogleApi _googleApi;
+
+  /// The client to interact with Firestore.
+  final FirestoreClient _firestoreClient;
+
+  /// The base URL of DV360.
+  final String _dv360BaseUrl;
```
cathyang-work, line 22: If it is a fixed string, we can just define it here.

server/test/run_rule_controller_test.dart
```diff
@@ -0,0 +1,91 @@
...
+      response = await runRuleController.runRule(scheduledRule.writeToBuffer());
+    });
+
+    test('calls FirestoreClient.getRule() with user ID', () async {
+      verify(mockFirestoreClient.getRule(userId, any));
+    });
+
+    test('calls FirestoreClient.getRule() with rule ID', () async {
+      verify(mockFirestoreClient.getRule(any, ruleId));
+    });
```
cathyang-work, line 88: I think it is find to join these 2 tests together 

```
 verify(mockFirestoreClient.getRule(userId, ruleId));
```

cathyang-work (changes requested):

server/test/run_rule_controller_test.dart
```diff
@@ -0,0 +1,91 @@
...
+
+class MockRule extends Mock implements Rule {}
+
+class TestRunRuleController extends RunRuleController {
+  TestRunRuleController(GoogleApi googleApi, FirestoreClient firestoreClient,
+      String aesKey, String baseUrl)
+      : super(googleApi, firestoreClient, aesKey, baseUrl);
+
+  @override
+  Rule getRule(proto.Rule ruleProto) {
```
gg-qinp: instead of overriding getRule(), can we let mockFirestoreClient.getRule() return a real rule object, then we don't have to test getRule().

server/lib/controller/run_rule_controller.dart
thu5: I was trying to use method dependency injection to make testing easier. I'll remove it, since I refactored to `await dv360Client.run(rule);`.

server/lib/controller/run_rule_controller.dart
thu5: Oops, it's not in this PR. It should be in `config.yaml`. I'll add it.

server/lib/controller/run_rule_controller.dart
thu5: Yes, that definitely works. I'll refactor it!

server/test/run_rule_controller_test.dart
thu5: Ok, that makes sense!

server/lib/controller/run_rule_controller.dart
thu5: It's actually not a fixed string. During testing, I changed it so that we use a mock DV360 server.

server/test/run_rule_controller_test.dart
thu5: Unfortunately, that doesn't completely work because the `DisplayVideo360Client` created in the method (which I can't mock) will try to run the rule and fail. Instead, I have removed the inheritance by mocking the DV360 server.

cathyang-work (approved):

gg-qinp (approved):

PR #109, opened 2020-08-21
Create a service class to interact with the reporting API

### LINKED ISSUE #106: Create a service class for reporting API

thu5:
This PR creates a service class to interact with the DV360 reporting API. It can currently get the CPM of a line item. Resolves #106.


server/lib/service/reporting.dart
```diff
@@ -0,0 +1,125 @@
...
+
+    // Creates the query for the line item.
+    final queryId = await _createLineItemQuery(
+        advertiserId, lineItemId, dataRange, metrics, groupBys);
+
+    // Get the URL of the report of the created query.
+    final url = await _getReportUrl(queryId);
+
+    try {
+      // Downloads the report and converts it into a map.
```
cathyang-work: Where is it gonna be download to?

server/test/reporting_test.dart
```diff
@@ -0,0 +1,187 @@
...
+    await mockReportingServer.close();
+  });
+
+  tearDown(() async {
+    mockReportingServer.clear();
+  });
+
+  group('Success case: getLineItemCpm()', () {
+    setUp(() async {
+      // Responds to the create query request with a [Query] containing the id.
```
cathyang-work: Is `[Query]` a class? Where is it defined?

server/test/reporting_test.dart
```diff
@@ -0,0 +1,187 @@
...
+    test('makes a request to the correct path to create the query', () async {
+      expect(
+          createQueryRequest.path.string, '/doubleclickbidmanager/v1.1/query');
+    });
+
+    test('makes a request with the correct body to create the query', () async {
+      final body = json.encode(await createQueryRequest.body.decode());
+      final expected = json.encode(createRequestBody);
+
+      expect(body, equals(expected));
```
cathyang-work: why do we need `equals`?

server/test/reporting_test.dart
```diff
@@ -0,0 +1,187 @@
...
+      expect(getReportRequest.path.string, '/$reportPath');
+    });
+
+    test('returns the correct CPM value', () async {
+      expect(returnValue, equals(cpm));
+    });
+  });
+
+  group('Failure case: getLineItemCpm()', () {
+    test('throws an ApiRequestError if error creating the query', () async {
```
cathyang-work: if [ it fails to] create the query. Ditto for the rest

server/lib/service/reporting.dart
gg-qinp: Ted@ correct me if I am wrong. I talked to Jin@ previously, the reporting API won't return data directly, instead it would store your query results into a CSV file in Google Cloud Storage, so we need to download and parse the CSV file to get exact data.

server/lib/service/reporting.dart
```diff
@@ -0,0 +1,125 @@
...
+  ///
+  /// Throws an [ApiRequestError] if there is an error with the reporting API.
+  Future<String> _getReportUrl(String queryId) async {
+    final query = await _api.queries.getquery(queryId);
+    return query.metadata.googleCloudStoragePathForLatestReport;
+  }
+
+  /// Extract the values in the report from [reportUrl] into a map.
+  ///
+  /// Throws an ArgumentError if report is not found or any data is missing.
```
gg-qinp, line 110: will the ArgumentError be throw by default from get(reportUrl) or we need to catch?

server/lib/service/reporting.dart
```diff
@@ -0,0 +1,125 @@
...
+    // Sends an HTTP GET request to the [reportUrl].
+    final response = await get(reportUrl);
+
+    // Converts the response body which is in CSV format into a list.
+    final rows = const CsvToListConverter()
+        .convert(response.body, eol: '\n', shouldParseNumbers: false);
+
+    // Converts the list into a map of keys and values.
+    final map =
+    Map.fromIterables(rows[0].cast<String>(), rows[1].cast<String>());
```
gg-qinp, line 131: is this format expected?

gg-qinp (approved):

server/lib/service/reporting.dart
cathyang-work: When I see "download", I think it is saving the file somewhere. But from in implementation of _reportToMap(), it seems that the csv is in the body of the GET response. So, would "retrieves" or "get" be more accurate?

server/lib/service/reporting.dart
```diff
@@ -0,0 +1,125 @@
...
+  /// Throws an ArgumentError if report is not found or any data is missing.
+  Future<Map<String, String>> _reportToMap(String reportUrl) async {
+    // Sends an HTTP GET request to the [reportUrl].
+    final response = await get(reportUrl);
+
+    // Converts the response body which is in CSV format into a list.
+    final rows = const CsvToListConverter()
+        .convert(response.body, eol: '\n', shouldParseNumbers: false);
+
+    // Converts the list into a map of keys and values.
```
cathyang-work, line 129: Can we add some comment to explain what the key and values is? If it is general, maybe give an example.

server/lib/service/reporting.dart
thu5: Yes, sorry that was a bad choice of wording on my part. "retrieves" makes more sense.

server/test/reporting_test.dart
thu5: Yes, it is defined [here](https://pub.dev/documentation/googleapis/latest/googleapis.doubleclickbidmanager.v1_1/Query-class.html) in the googlesapi package. I'll make the comment clearer.

server/lib/service/reporting.dart
thu5: The ArgumentError will be thrown by get(reportUrl). We shouldn't catch it yet (if the report is somehow missing). It should propagate up a bit higher until we record the run history for each rule.

server/lib/service/reporting.dart
thu5: Yes, this should be the expected format of the report. If no data is there, an argument error is thrown.

server/lib/service/reporting.dart
thu5: Sure!

cathyang-work: Have you updated the comment?

thu5: I added it to the [doc comment](https://github.com/googleinterns/dv360-excel-plugin/blob/b620fadfc330aa5de033551683fe4280a6f12a34/server/lib/service/reporting.dart#L112) for _retrieveReport().

cathyang-work (approved):

PR #115, opened 2020-08-27
Add service class in client to interact with rule server

### LINKED ISSUE #74: Add rules service

thu5:
This PR adds a service class in the client to interact with the rule server. Closes #74.

### DIFF
--- a/lib/src/service/rule_service.dart
+++ b/lib/src/service/rule_service.dart
@@ -1,0 +1,41 @@
+import 'dart:async';
+
+import 'package:angular/angular.dart';
+import 'package:http/http.dart';
+
+import '../data_model/create_user_request.pb.dart';
+import '../data_model/rule.pb.dart';
+
+/// A class to interact with the Rules Builder server.
+@Injectable()
+class RuleService {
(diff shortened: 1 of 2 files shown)


lib/src/service/rule_service.dart
```diff
@@ -0,0 +1,41 @@
...
+
+import '../data_model/create_user_request.pb.dart';
+import '../data_model/rule.pb.dart';
+
+/// A class to interact with the Rules Builder server.
+@Injectable()
+class RuleService {
+  /// The base URL of the Rules Builder server.
+  static const serverUrl =
+      'https://rules-server-dot-spreadsheet-dv360-plugin.uc.r.appspot.com';
```
gg-qinp, line 14: it is better to put endpoint in the yaml file.

gg-qinp (approved):

lib/src/service/rule_service.dart
thu5: On the client side, there isn't a yaml file or config file.

cathyang-work (approved):

PR #117, opened 2020-08-27
Add scope to proto

### LINKED ISSUE #116: Add scope to proto

thu5:
This PR adds a Scope message to `rule.proto` so that the scope of the rule is decoupled from the action. The reason for this change is to allow conditions to also access scope data without having to reach into an action. Closes #116.

Tests will be added later (#118).

### DIFF
--- a/server/lib/model/scope.dart
+++ b/server/lib/model/scope.dart
@@ -1,0 +1,41 @@
+import 'package:fixnum/fixnum.dart';
+
+import '../proto/rule.pb.dart' as proto;
+
+/// A class that represents the target entities for the rule.
+class Scope {
+  /// A list of target DV360 entities.
+  List<Target> targets = [];
+
+  /// Creates an instance of a class that implements [Scope] from a proto.
+  ///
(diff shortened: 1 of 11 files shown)


server/lib/model/scope.dart
```diff
@@ -0,0 +1,41 @@
+import 'package:fixnum/fixnum.dart';
+
+import '../proto/rule.pb.dart' as proto;
+
+/// A class that represents the target entities for the rule.
+class Scope {
+  /// A list of target DV360 entities.
+  List<Target> targets = [];
```
cathyang-work, line 8: Where is Target defined?

thu5: [Here](https://github.com/googleinterns/dv360-excel-plugin/blob/133c37d95e0216c8c291f8fa28b41875f0ab7f53/server/lib/model/scope.dart#L29) it is. It's just a marker interface.

cathyang-work (approved):

gg-qinp (approved):

PR #123, opened 2020-08-27
Set up run history

### LINKED ISSUE #122: Add run history for rules

thu5:
This PR adds a method to FirestoreClient to add the run history, and modifies DisplayVideo360Client's `run()` in order to log the performed rule. Resolves #122.

Will add tests later (#124).

### DIFF
--- a/server/lib/controller/run_rule_controller.dart
+++ b/server/lib/controller/run_rule_controller.dart
@@ -57,10 +57,11 @@ class RunRuleController extends ResourceController {
     final userClient = await _googleApi.getUserAccountClient(refreshToken);
 
     // Creates a DV360 client using the user's authenticated client.
-    final dv360 = DisplayVideo360Client(userClient, _dv360BaseUrl);
+    final dv360 =
+        DisplayVideo360Client(userClient, _firestoreClient, _dv360BaseUrl);
 
     // Runs the rule using the DV360 client.
-    await dv360.run(rule);
+    await dv360.run(rule, userId, ruleId);
 
     return Response.ok({});
   }
--- a/server/lib/service/dv360.dart
+++ b/server/lib/service/dv360.dart
@@ -3,9 +3,9 @@ import 'dart:async';
 import 'package:fixnum/fixnum.dart';
 import 'package:googleapis/displayvideo/v1.dart';
 import 'package:http/http.dart';
-import 'package:server/service/firestore.dart';
 
 import '../model/rule.dart';
+import '../service/firestore.dart';
 
 /// A class that wraps around Display & Video 360.
 class DisplayVideo360Client {
@@ -37,13 +37,11 @@ class DisplayVideo360Client {
       await rule.action.run(this);
     } on ApiRequestError catch (e) {
       // If there is an API error, return the message returned by the API.
-      await _firestoreClient.addRun(userId, ruleId,
+      return await _firestoreClient.addRun(userId, ruleId,
           isSuccess: false, message: e.message);
-      return;
     } catch (e) {
       // If there is another kind of exception, do not include the message.
-      await _firestoreClient.addRun(userId, ruleId, isSuccess: false);
-      return;
+      return await _firestoreClient.addRun(userId, ruleId, isSuccess: false);
     }
     // Logs the successful run of the rule.
     await _firestoreClient.addRun(userId, ruleId, isSuccess: true);
--- a/server/lib/service/dv360.dart
+++ b/server/lib/service/dv360.dart
@@ -37,13 +37,13 @@ class DisplayVideo360Client {
       await rule.action.run(this);
     } on ApiRequestError catch (e) {
       // If there is an API error, return the message returned by the API.
-      return await _firestoreClient.addRun(userId, ruleId,
-          isSuccess: false, message: e.message);
+      return await _firestoreClient.logRunHistory(userId, ruleId, false,
+          message: e.message);
     } catch (e) {
       // If there is another kind of exception, do not include the message.
-      return await _firestoreClient.addRun(userId, ruleId, isSuccess: false);
+      return await _firestoreClient.logRunHistory(userId, ruleId, false);
     }
     // Logs the successful run of the rule.
-    await _firestoreClient.addRun(userId, ruleId, isSuccess: true);
+    await _firestoreClient.logRunHistory(userId, ruleId, true);
   }
 }
--- a/server/lib/service/dv360.dart
+++ b/server/lib/service/dv360.dart
@@ -41,7 +41,13 @@ class DisplayVideo360Client {
           message: e.message);
     } catch (e) {
       // If there is another kind of exception, do not include the message.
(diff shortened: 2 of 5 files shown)


server/lib/controller/run_rule_controller.dart
```diff
@@ -57,10 +57,11 @@ class RunRuleController extends ResourceController {
     final userClient = await _googleApi.getUserAccountClient(refreshToken);
 
     // Creates a DV360 client using the user's authenticated client.
-    final dv360 = DisplayVideo360Client(userClient, _dv360BaseUrl);
+    final dv360 =
+        DisplayVideo360Client(userClient, _firestoreClient, _dv360BaseUrl);
 
     // Runs the rule using the DV360 client.
-    await dv360.run(rule);
+    await dv360.run(rule, userId, ruleId);
```
cathyang-work, line 64: Is `ruleId` also existing in `rule`?

server/lib/service/firestore.dart
```diff
@@ -63,6 +67,29 @@ class FirestoreClient {
     );
   }
 
+  /// Logs [isSuccess] and optional [message] of a performed rule with [ruleId].
+  ///
+  /// The [userId] is the `sub` claim of the user's Google ID token.
+  /// See: https://developers.google.com/identity/protocols/oauth2/openid-connect#an-id-tokens-payload
+  ///
+  /// Throws an [ApiRequestError] if Firestore API returns an error.
+  Future<void> addRun(String userId, String ruleId,
```
cathyang-work: Suggest to rename the method name to `logRunHistory` to be more clear.

server/lib/service/firestore.dart
```diff
@@ -63,6 +67,29 @@ class FirestoreClient {
...
+      {@required bool isSuccess, String message = ''}) async {
```
cathyang-work: I think moving `bool isSuccess` out of the `{}` would indicate this param is required.

server/lib/service/firestore.dart
```diff
@@ -12,6 +13,9 @@ class FirestoreClient {
   /// The name of the collection of rules in Firestore.
   static const rulesName = 'rules';
 
+  /// The name of the collection of the run history in Firestore.
+  static const runHistoryName = 'runs';
```
cathyang-work: rename it to "runHistory" or 'runLog' ?

server/lib/service/dv360.dart
```diff
@@ -27,8 +31,19 @@ class DisplayVideo360Client {
...
+  Future<void> run(Rule rule, String userId, String ruleId) async {
+    try {
+      await rule.action.run(this);
+    } on ApiRequestError catch (e) {
+      // If there is an API error, return the message returned by the API.
+      return await _firestoreClient.addRun(userId, ruleId,
+          isSuccess: false, message: e.message);
+    } catch (e) {
+      // If there is another kind of exception, do not include the message.
+      return await _firestoreClient.addRun(userId, ruleId, isSuccess: false);
```
cathyang-work: Why we don't log msg is it is not `ApiRequestError`?

server/lib/controller/run_rule_controller.dart
thu5: The `Rule` model currently doesn't store the rule ID, only the `Rule` proto does. Should I add it to the Rule model?

server/lib/service/firestore.dart
thu5: Okay!

server/lib/service/firestore.dart
thu5: Ok, that makes sense. I did it this way because one of Aqueduct's default lint rules is to [avoid positional boolean parameters](https://dart-lang.github.io/linter/lints/avoid_positional_boolean_parameters.html).

server/lib/service/dv360.dart
thu5: Our target users are not necessarily developers, so the messages we log should be user-friendly, actionable and understandable. We can expect this for DV360 API error messages, but probably not for lower level exception messages. Also, there might be security issues if we just directly report the raw exception messages.

cathyang-work: Clear explanation! Maybe you can add a comment to the catch block?

server/lib/service/firestore.dart
cathyang-work: If that is the case, you can keep what you had before.

server/lib/controller/run_rule_controller.dart
cathyang-work: I don't have a strong opinion.

If you decide the add `ruleId` to Rule model, you can remove the `ruleId` in `DV360Client.run()`. 

If not, I see that `Rule` model only have one field: `Action`. Do you plan to add other fields in the future? If not, I think you can just use DV360Client.run(ruleProto...) and convert the Action proto to Action object inside `DV360Client.run()`, instead of converting the Rule Proto to an Rule object and only get the Action object from the Rule object.

server/lib/service/dv360.dart
```diff
@@ -27,8 +31,19 @@ class DisplayVideo360Client {
...
+  Future<void> run(Rule rule, String userId, String ruleId) async {
+    try {
+      await rule.action.run(this);
+    } on ApiRequestError catch (e) {
+      // If there is an API error, return the message returned by the API.
+      return await _firestoreClient.logRunHistory(userId, ruleId, false,
+          message: e.message);
+    } catch (e) {
+      // If there is another kind of exception, do not include the message.
+      return await _firestoreClient.logRunHistory(userId, ruleId, false);
```
gg-qinp: nit: you can put a message like `internal error encountered` here

server/lib/service/dv360.dart
```diff
@@ -27,8 +31,19 @@ class DisplayVideo360Client {
...
-  /// Runs the rule to manipulate DV360 line items.
-  Future<void> run(Rule rule) async {
-    await rule.action.run(this);
+  /// Runs the rule to manipulate DV360 line items and logs the result.
+  Future<void> run(Rule rule, String userId, String ruleId) async {
+    try {
+      await rule.action.run(this);
+    } on ApiRequestError catch (e) {
+      // If there is an API error, return the message returned by the API.
+      return await _firestoreClient.logRunHistory(userId, ruleId, false,
```
gg-qinp: seems no test coverage for these newly added code? I am ok you add it in later pr.

gg-qinp (approved):

server/lib/controller/run_rule_controller.dart
thu5: The `Rule` model will also have a `condition` field in a later PR. So I guess, just keep the same for now?

server/lib/service/dv360.dart
thu5: Sure, that sounds good!

server/lib/service/dv360.dart
thu5: Yes, I'll add it in a later PR.

cathyang-work (approved):

PR #125, opened 2020-09-02
Add rule creation UI

### LINKED ISSUE #70: Create a basic UI for rules builder
Basic functionality only, styling can be done later.

thu5:
This PR adds the UI for rule creation. Closes #70.

output


lib/src/rule_creator_component.dart
```diff
@@ -0,0 +1,170 @@
...
+
+@Component(
+  selector: 'rule-creator',
+  templateUrl: 'rule_creator_component.html',
+  directives: [coreDirectives, formDirectives, BsInput, bsAccordionDirectives],
+  providers: [ClassProvider(RuleService)],
+)
+class RuleCreatorComponent {
+  final RuleService _ruleService;
+  final CredentialService _credentialService;
```
gg-qinp, line 32: why this is not within `providers` above

gg-qinp (approved):

lib/root_component.html
```diff
@@ -61,7 +61,16 @@ <h6 *ngIf="showSideloadMessage" class="secondary-msg" debugId="sideload-msg">
...
         <!-- rules builder-->
         <bs-tabx header="{{rulesBuilderTabName}}">
-            <p>Coming soon</p>
+            <div class="jumbotron d-flex align-items-center">
+                <div class="container-fluid">
+                    <!-- rule component-->
+                    <rule-creator></rule-creator>
+                    <!-- sign off button-->
+                    <div class="d-flex justify-content-end">
+                        <small><a debugId="sign-off-btn" (click)="onClick()" href="#">{{signOutButtonName}}</a></small>
```
cathyang-work: Why do we use `href="#"` here? Is it going to be replaced by something else later?

cathyang-work (approved):

lib/src/rule_creator_component.dart
thu5: Oops, it should be! I'll fix it

lib/root_component.html
thu5: That code is reused from the query builder, but no it doesn't need to be there. I'll remove it.

PR #126, opened 2020-09-03
Add rule list view component

### LINKED ISSUE #76: Create rule list component
Basic functionality.

thu5:
This PR adds the rule list component to display all the user's rules. Also did some refactoring. Closes #76.

list


lib/src/rule_list_component.html
```diff
@@ -0,0 +1,9 @@
+<div class="w-auto overflow-auto" style="max-height: 60vh">
+    <a *ngFor="let t of rules" href="#" class="list-group-item list-group-item-action flex-column align-items-start">
```
cathyang-work: What does `t` mean? Maybe `r` or `rule`?

server/lib/controller/rule_controller.dart
```diff
@@ -51,4 +51,26 @@ class RuleController extends ResourceController {
 
     return Response.ok({});
   }
+
+  /// Retrieves the user's rules from Firestore.
+  ///
+  /// Throws [ArgumentError] if request doesn't contain an Authorization header.
```
cathyang-work: I guess we need `an` for consistency.

server/lib/service/firestore.dart
```diff
@@ -93,6 +140,24 @@ class FirestoreClient {
...
+  /// The [pageSize] is the maximum number of documents to return. The
+  /// [pageToken] is the token used to retrieve the next page.
+  ///
+  /// Throws an [ApiRequestError] if Firestore API returns an error.
+  Future<ListDocumentsResponse> _listDocuments(String parent, String collection,
+      {int pageSize, String pageToken}) async {
+    final documents = await _api.projects.databases.documents.list(
+      'projects/$_projectId/databases/$_databaseId/documents$parent',
+      collection,
+      pageSize: pageSize,
```
cathyang-work: How is  the `pageSize` gonna be used in UI? If we do need the `pageSize`? Can we rename it to `maxPageSize`?

lib/src/rule_list_component.html
thu5: It doesn't mean anything, I was just lazy 😅. `rule` is good!

server/lib/controller/rule_controller.dart
thu5: Sure!

server/lib/service/firestore.dart
thu5: Currently I don't plan to use pageSize in the UI, but pagination could be possible in the future. `maxPageSize` is a better name though, I'll change it.

cathyang-work (approved):

lib/src/service/rule_service.dart
```diff
@@ -0,0 +1,56 @@
...
+
+import '../data_model/create_user_request.pb.dart';
+import '../data_model/get_rules_response.pb.dart';
+import '../data_model/rule.pb.dart';
+
+/// A class to interact with the Rules Builder server.
+@Injectable()
+class RuleService {
+  /// The base URL of the Rules Builder server.
+  static const serverUrl =
```
gg-qinp: seems the web server is calling app engine server to get rules, curious why we don't call _firestoreClient.getUserRule() directly within this file.

gg-qinp (approved):

lib/src/service/rule_service.dart
thu5: Currently only the web server interacts with Firestore using the default app engine service account. This is because the server stores all the users' rules, and we don't give the client the app engine service account keys. 

If each user stored their own rules, I guess we could have the Excel client directly call Firestore. But we would have to change the design significantly.

PR #127, opened 2020-09-07
Handle requests to get run history for rule

### LINKED ISSUE #122: Add run history for rules

thu5:
This PR allows the server to handle requests to get the run history for a rule. Resolves #122.


get_run_history.proto
```diff
@@ -0,0 +1,24 @@
...
+
+// A message that represents the response for the run history.
+message GetRunHistoryResponse {
+  // required
+  // The list of run histories.
+  repeated Run history = 1;
+}
+
+// A message that represents a single execution of a rule.
+message Run {
```
cathyang-work: Maybe rename 'Run' to something like 'RunLog', 'RunEntry'? I think it will be more clear when ppl are reading the `Run` proto in another file. Also, do you need to include the ruleId too?

cathyang-work (approved):

get_run_history.proto
thu5: Yes, that makes sense. `ruleId` isn't necessary because the run history is a subcollection of each rule.

gg-qinp (approved):

Issue #77, opened 2020-07-25
Create rule detail component


thu5:
Basic functionality.

PR #128, opened 2020-09-08
Add rule details view with run history

### LINKED ISSUE #122: Add run history for rules

thu5:
This PR adds the run details UI which includes details about the rule and the run history. Resolves #122 and #77.

success  failure


lib/src/rule_component.html
```diff
@@ -0,0 +1,14 @@
+<div class="mb-3" *ngIf="view == View.listView">
+    <div class="flex-column">
+        <rule-list (detailedRule)="changeDetailedRule($event)" class="mb-5"></rule-list>
+        <button type="button" (click)="toCreateView()" class="mt-3 btn btn-success">Create Rule</button>
+    </div>
+</div>
+<div *ngIf="view == View.createView">
+    <rule-creator></rule-creator>
+    <button type="button" (click)="toListView()" class="btn btn-secondary">Back</button>
+</div>
```
gg-qinp, line 10: does Git have a delta diff view which omits the diff from previous RP?

lib/src/service/rule_service.dart
```diff
@@ -0,0 +1,70 @@
...
+      'Authorization': 'Bearer $idToken'
+    };
+    final response = await get(rulesEndpoint, headers: headers);
+    return GetRulesResponse.fromBuffer(response.bodyBytes).rules;
+  }
+
+  Future<GetRunHistoryResponse> getRunHistory(
+      String idToken, String ruleId) async {
+    final headers = {
+      'Content-Type': 'application/x-protobuf',
```
gg-qinp: consider make this a static function instead of defining it repeatedly.

gg-qinp (approved):

lib/src/service/rule_service.dart
thu5: Ok, sure

lib/src/rule_component.html
thu5: I don't think so :/

cathyang-work (approved):

PR #134, opened 2020-09-08
Add one-time schedule

### LINKED ISSUE #133: Add one-time schedule

thu5:
This PR allows users to add one-time schedules for rules. Closes #133.

Screen Shot 2020-09-08 at 2 03 03 PM


cathyang-work (commented):
Can you add a screenshot?

thu5:
Sure

cathyang-work (approved):

lib/src/rule_creator_component.dart
```diff
@@ -130,7 +130,16 @@ class RuleCreatorComponent {
...
-            ..cronExpression = _cronExpression(month, day, hour, minute));
+            ..cronExpression =
+                _repeatingCronExpression(month, day, hour, minute));
+        break;
+      case Schedule_Type.ONE_TIME:
+        rule.schedule = Schedule()
+          ..type = scheduleTypes[scheduleType]
+          ..timezone = timezone
+          ..oneTimeParams = (Schedule_OneTimeParams()
+            ..cronExpression = '$minute $hour $day $month *'
```
gg-qinp, line 141: [No Action Required]
1. theoretically we should make sure the time is not a past time,but it is fine for now.
2. current special handling for one time task does not scale for task with more repeating times. Think about how Google Calendar handle repeated tasks, we can store multiple entries with timestamp in a table,  then we can maintain a background job to query the table periodically and trigger the task.

gg-qinp (approved):

lib/src/rule_creator_component.dart
thu5: Good points, I'll talk about it in my presentation :) I can handle (1) if I have time though.

PR #138, opened 2020-09-09
Add change bid strategy action

### LINKED ISSUE #136: Add change bidding strategy action

thu5:
This PR adds the ability for users to change the bidding strategy of line items to:

- FixedBidStrategy
    - Requires bid amount.
- MaximizeSpendBidStrategy
    - Requires performance goal type.
- PerformanceGoalBidStrategy
    - Requires performance goal type and performance goal amount.

The 'Viewable Impressions' goal option maps to 'BIDDING_STRATEGY_PERFORMANCE_GOAL_TYPE_AV_VIEWED' when the strategy is maximize spend and maps to 'BIDDING_STRATEGY_PERFORMANCE_GOAL_TYPE_VIEWABLE_CPM' when the strategy is performance goal.

I decided not to implement maxAverageCpmBidAmount and customBiddingAlgorithm. 

Closes #136.


rule.proto
```diff
@@ -54,6 +56,60 @@ message ChangeLineItemStatusParams {
...
+}
+
+// A message representing the parameters for fixed bid strategy.
+message FixedBidStrategyParams {
+  // required
+  // The fixed bid amount (micros) in the advertiser's currency.
+  optional int64 bid_amount_micros = 1;
+}
+
+enum BiddingStrategyPerformanceGoalType {
```
cathyang-work, line 88: I am curious how you find these BiddingStrategyPerformanceGoal?

server/lib/service/dv360.dart
```diff
@@ -31,6 +31,36 @@ class DisplayVideo360Client {
...
+            (MaximizeSpendBidStrategy()..performanceGoalType = goal);
+        break;
+      case 'PERFORMANCE_GOAL':
+        request.bidStrategy.performanceGoalAutoBid =
+            (PerformanceGoalBidStrategy()
+              ..performanceGoalType = goal
+              ..performanceGoalAmountMicros = goalAmount.toString());
+        break;
+      default:
+        throw UnimplementedError('$strategy is not a supported strategy.');
```
cathyang-work: If there are not more strategy to be added, maybe use UnsupportedError like other places? Unless you plan to add more cases in here.

server/lib/model/action.dart
```diff
@@ -61,3 +65,151 @@ class ChangeLineItemStatusAction implements Action {
...
+            // If maximize spend strategy:
+            goal = 'BIDDING_STRATEGY_PERFORMANCE_GOAL_TYPE_AV_VIEWED';
+          }
+          if (_biddingStrategy == 3) {
+            // If performance goal strategy
+            goal = 'BIDDING_STRATEGY_PERFORMANCE_GOAL_TYPE_VIEWABLE_CPM';
+          }
+          break;
+        default:
+          throw UnsupportedError(
```
cathyang-work: Maybe `$_performanceGoal is an unsopported performance goal.`? What is an example for `$_performanceGoal`?

server/lib/model/action.dart
```diff
@@ -61,3 +65,151 @@ class ChangeLineItemStatusAction implements Action {
...
+      default:
+        throw UnsupportedError(
+            '$_biddingStrategy is an invalid value for a bidding strategy.');
+    }
+    return action;
+  }
+
+  /// Changes the bidding strategy of the line item using the DV360 client.
+  @override
+  Future<void> run(DisplayVideo360Client client, Target target) async {
```
cathyang-work: I think Action.dart is used for mapping, like converting the proto to POJO, which that can be used to send request to DV3Client. What the `run` method wants to achieve seems to be OOO of the purpose of this file. Maybe we should move the `run` method  in another class like dv360.dart.

So in dv360.dart, we can add a method like this:

```
run(target, action){
    switch(action) {
        case ChangeLineItemStatusAction:
           changeLineItemStatus(...);
        case ChangeLineItemBiddingStrategyAction:
           changeLineItemBiddingStrategy(...); 
        ...
   }
}
```

rule.proto
thu5: I looked at the DV360 Line Item Editor UI and referenced the [Public API documentation](https://developers.google.com/display-video/api/reference/rest/v1/BiddingStrategy#BiddingStrategyPerformanceGoalType). In the end, I only included a subset of the goal types.

server/lib/service/dv360.dart
thu5: That makes sense! I don't plan on adding more strategies here.

server/lib/model/action.dart
thu5: `_performanceGoal` is an integer value corresponding to the enum value of `BiddingStrategyPerformanceGoalType`. So 1, 2, 3, or 0 if unspecified.

server/lib/model/action.dart
thu5: Ok, yeah, that makes sense. I'll refactor.

server/lib/model/action.dart
cathyang-work: Does user know what these 1, 2,3 are mapped to? Can we return the name instead of number to make the error msg more helpful?

thu5: Yes, sure!

cathyang-work (approved):

server/lib/model/action.dart
```diff
@@ -45,19 +42,101 @@ class ChangeLineItemStatusAction implements Action {
...
   }
+}
 
-  /// Changes the status of the line item using the DV360 client.
-  /// TODO(@thu5): Record the return status and implement run history.
-  @override
-  Future<void> run(DisplayVideo360Client client, Target target) async {
-    final lineItemTarget = target as LineItemTarget;
+/// A class that represents changing the line item bidding strategy.
+class ChangeLineItemBiddingStrategyAction implements Action {
```
gg-qinp: nit: As we add more and more rules, this file and proto will become super large. it is better to split them into individual file by rule.

server/lib/model/action.dart
```diff
@@ -45,19 +42,101 @@ class ChangeLineItemStatusAction implements Action {
...
+  /// The integer value of the bidding strategy.
+  int biddingStrategy;
```
gg-qinp: wondering can we use enum here, so later the switch case can use enum, which has better readability.

gg-qinp (approved):

server/lib/model/action.dart
thu5: Ok, yes, that makes sense. I'll split them up.

server/lib/model/action.dart
thu5: Thank you! That makes it a LOT more readable.

PR #140, opened 2020-09-10
Add duplicate line item(s) action

### LINKED ISSUE #137: Add duplicate line items action

thu5:
This PR adds the ability to duplicate line items. Resolves #137.


server/lib/service/dv360.dart
```diff
@@ -31,11 +34,58 @@ class DisplayVideo360Client {
...
   /// Runs the rule to manipulate DV360 line items and logs the result.
   Future<void> run(Rule rule, String userId, String ruleId) async {
     for (final target in rule.scope.targets) {
       try {
-        await rule.action.run(this, target);
+        switch (rule.action.runtimeType) {
+          case ChangeLineItemStatusAction:
+            await runStatusAction(rule.action, target);
+            break;
+          case DuplicateLineItemAction:
```
cathyang-work: Make sure after you merge this pr, the big strategy action is still there.

cathyang-work (approved):

server/lib/service/dv360.dart
```diff
@@ -31,11 +34,58 @@ class DisplayVideo360Client {
         updateMask: 'entityStatus');
   }
 
+  /// Duplicates the line item to [insertionOrderIdDestination].
+  ///
+  /// The [advertiserIdDestination] is the ID of the destination advertiser.
+  ///
+  /// Throws an [ApiRequestError] if API returns an error.
+  Future<void> duplicateLineItem(Int64 advertiserId, Int64 lineItemId,
```
gg-qinp, line 44: fyi: a complete line item contains targeting which is a major part. but this requires much extra effort. so we are good now.

server/lib/service/dv360.dart
```diff
@@ -31,11 +34,58 @@ class DisplayVideo360Client {
...
+    source
+      ..insertionOrderId = insertionOrderIdDestination.toString()
+      ..entityStatus = 'ENTITY_STATUS_DRAFT'
+      ..advertiserId = null
+      ..campaignId = null
+      ..lineItemId = null
+      ..updateTime = null
+      ..partnerCosts = null
+      ..name = null
+      ..flight.dateRange = null
```
gg-qinp: why not reuse the flight from source line item?

gg-qinp (approved):

server/lib/service/dv360.dart
thu5: When I was writing it, I wasn't sure if the flight date should be inherited from the destination parent insertion order. When it does inherit, the dateRange is output only (which is why I set it to null).

So we should make the new flightDateType LINE_ITEM_FLIGHT_DATE_TYPE_CUSTOM and have the same dateRange as the source?

gg-qinp: i see. actually although it is input only, you can still set it, the server just ignore the input, :p, the better way is to add check `If isInherited`, make dateRange = null.

thu5: Ok gotcha!

PR #142, opened 2020-09-15
Make condition optional and improve symbols

thu5:
This PR makes condition optional and improves the symbols used in the UI (<= to ≤).

### DIFF
--- a/lib/src/rule_creator_component.html
+++ b/lib/src/rule_creator_component.html
@@ -267,13 +267,13 @@
         <bs-accordion-panel heading="Condition Parameters (Optional)" (click)="firstPanelOpen = false">
             <div class="form-group">
                 <label class="parameter-name">Condition Type</label>
-                <select class="form-control" #field="ngForm" [(ngModel)]="conditionType" ngControl="conditionType" required>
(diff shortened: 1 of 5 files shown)


lib/src/rule_creator_component.html
```diff
@@ -267,21 +267,21 @@
...
+                <select class="form-control" #field="ngForm" [(ngModel)]="relation" ngControl="relation">
                     <option *ngFor="let t of relationTypes.keys" [value]="t">{{t}}</option>
                 </select>
             </div>
             <div class="form-group">
                 <bs-input class="is-valid"
                           label="Value"
                           [(ngModel)]="value"
-                          pattern="^(?:[1-9]\d*|0)?(?:\.\d+)?$"
+                          pattern="((^$)|(^(?:[1-9]\d*|0)?(?:\.\d+)?$))"
```
gg-qinp: curious what is `(^$)` matching for?

gg-qinp (approved):

cathyang-work (approved):

lib/src/rule_creator_component.html
thu5: My mistake, it doesn't do anything.

PR #144, opened 2020-09-15
Allow optional setup of Identity-Aware proxy

thu5:
This PR allows the optional use of [Identity-Aware Proxy](https://cloud.google.com/iap) (IAP) to guard access to the server. It also refactors getting the encoded id token from the request, by creating a new utils function.

### DIFF
--- a/server/lib/controller/id_token_validator.dart
+++ b/server/lib/controller/id_token_validator.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 
 import 'package:jose/jose.dart';
 import 'package:aqueduct/aqueduct.dart';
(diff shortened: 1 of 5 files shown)


gg-qinp (approved):
Seems the IAP is used to protect your server in app engine. Do we need this change if you choose to not enable IAP for you application?

cathyang-work (approved):

thu5:
> Seems the IAP is used to protect your server in app engine. Do we need this change if you choose to not enable IAP for you application?

We don't need this change we don't enable IAP. But this change allows both with-IAP and without-IAP.

PR #145, opened 2020-09-18
Move button to rule component and use session storage for ID token

thu5:
This PR moves the sign in button from the rule creator view to the list view and uses session storage instead of local storage for the ID token. This is because the ID token expires after 1 hour, and there is limited time to implement a better solution.

### DIFF
--- a/lib/src/rule_component.html
+++ b/lib/src/rule_component.html
@@ -1,7 +1,9 @@
 <div class="mb-3" *ngIf="view == View.listView">
     <div class="flex-column">
         <rule-list (detailedRule)="changeDetailedRule($event)" class="mb-5"></rule-list>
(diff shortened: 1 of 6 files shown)


gg-qinp (approved):

lib/src/rule_component.html
```diff
@@ -1,7 +1,9 @@
 <div class="mb-3" *ngIf="view == View.listView">
     <div class="flex-column">
         <rule-list (detailedRule)="changeDetailedRule($event)" class="mb-5"></rule-list>
-        <button type="button" (click)="toCreateView()" class="mt-3 btn btn-success">Create Rule</button>
+
```
cathyang-work: do you mean to have the extra line?

cathyang-work (approved):

lib/src/rule_component.html
thu5: Oops, mistake

PR #148, opened 2020-09-19
Fix client-side and server-side tests

thu5:
This PR fixes the client-side and server-side tests.


server/test/action_test.dart
```diff
@@ -44,15 +36,5 @@ void main() {
 
       expect(actionProto, equals(validActionProto));
     });
-
-    test('run() changes the line item status using the client', () async {
-      final status = 'ENTITY_STATUS_'
```
cathyang-work, line 49: Why do you delete this test?

thu5: The run method got moved to `dv360.dart` but I forgot to delete it here.

cathyang-work (approved):

